PR #8, opened 2019-12-28
Wrap a function's return type in a block

cgrushko:
Summary:
See new tests.

The handling of `longFunctionNoBlock` isn't ideal:

```
fun longFunctionNoBlock():
    Int =
    1234567 + 1234567

```

but I think it's a move in the right direction.

Differential Revision: D19245784


hick209:
I've reviewed the diff in Phabricator, @cgrushko

Issue #22, opened 2020-02-13
Space after link at end of sentence in KDoc


sgrimm:
The handling of links in the middle of sentences is much better now, but the new behavior introduced one quirk.

```
/**
 * This is a comment that links to [AnotherClass].
 */
```

gets formatted as

```
/**
 * This is a comment that links to [AnotherClass] .
 */
```

cgrushko:
cc @strulovich

strulovich:
This was fixed in https://github.com/facebookincubator/ktfmt/commit/c57cb64056325a76882a17fb489d7ae48ee72368#diff-1930e9b21d36d1f420b9a81b660b4b44

hick209:
We should close this issue, @cgrushko

PR #25, opened 2020-03-03
Honor original spacing after links

sgrimm:
There are legitimate reasons to want or not want whitespace after links in KDoc comments, so try to preserve the original intent but still clean up excessive spacing.

Fixes issue #22.


cgrushko:
@strulovich, can you take this?

cgrushko:
@sgrimm, does this cover removing trailing spaces before newlines, if there's a line wrap?

strulovich:
Sorry, I only saw this after I solved this.
I went with a different solution in the end that uses a specific markdown token, and it more strict - some tokens will be without space and some with, unrelated to the spaces before. So words will be spaced, and punctuation will not be separated.
It should be merged soon.
Thanks for the PR, and I'm honestly sorry we didn't fix this earlier on. It took more than it should have.

PR #34, opened 2020-05-05
feat: use official Kotlin token constants where possible

RanadeepPolavarapu:
Just adds a small nice-to-have, it uses official Kotlin tokens constants from the lexer.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/KotlinInputAstVisitor.kt
@@ -137,7 +137,7 @@ class KotlinInputAstVisitor(
     builder.block(ZERO) {
       visitFunctionLikeExpression(
           function.modifierList,
-          "fun",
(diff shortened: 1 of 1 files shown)


cgrushko:
@RanadeepPolavarapu, thank you!!!

I have a concern it'll make it harder to read the code. Right now, `")"` pops into my eye, whereas `KtTokens.LPAREN.Value` does less so. What do you think?

cc @strulovich

RanadeepPolavarapu:
@cgrushko Yes, I definitely agree. When I was making the change I thought this is inconvenient for the developers as we don't get direct visibility into the underlying token. My only though process was that this may be a safer approach to future typos.

strulovich:
This was worth checking and it did cross my mind, but looking at the result, I think the explicit token is clearer and shorter. 

My usual thinking on when to use constants are based on:
- Performance - I don't expect a meaningful change around this
- Changes in the constant - I don't expect Kotlin to change any of these values, so I think we are on safe ground for that.

Thanks for taking the time

PR #46, opened 2020-05-29
Fix a small README error

plumpy:


cgrushko:
@plumpy thanks, I sent a diff to fix this.

PR #53, opened 2020-07-14
Add static function to represent dropbox style option in formatting options

franvis:
Add a public api to represent dropbox style formatting options in order to avoid the duplication of formatting logic from consumers of ktfmt such as in: https://github.com/diffplug/spotless/pull/642


franvis:
Hey @cgrushko is it possible for you to do a `0.16` release to start using this? 😄

cgrushko:
@franvis  done :)

PR #62, opened 2020-09-29
chore: bump `intellij.version` for new EAP

### LINKED ISSUE #61: Plugin incompatible with AS 4.0.1

jef:
### Description

Fixes #61 

Solves issue where not compatible with latest EAP

image

image

### DIFF
--- a/ktfmt_idea_plugin/build.gradle.kts
+++ b/ktfmt_idea_plugin/build.gradle.kts
@@ -19,7 +19,7 @@ plugins {
     java
 }
 
-val ktfmtVersion = "0.19"
+val ktfmtVersion = "0.18"
 
 group = "com.facebook"
 version = "1.1-SNAPSHOT.$ktfmtVersion"
@@ -41,9 +41,11 @@ dependencies {
 
 // See https://github.com/JetBrains/gradle-intellij-plugin/
 intellij {
-    version = "2020.3"
+    version = "LATEST-EAP-SNAPSHOT"
 }
-tasks.getByName<org.jetbrains.intellij.tasks.PatchPluginXmlTask>("patchPluginXml") {
-    changeNotes("""
-      First version.""")
+
+tasks {
+    patchPluginXml {
+        sinceBuild("203")
+    }
 }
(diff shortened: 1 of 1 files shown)


cgrushko:
Thanks!
Does it mean that the minimum required version is 2020.3?

jef:
> Thanks!
> Does it mean that the minimum required version is 2020.3?

Well I think usually we want to do something like this:

```diff
intellij {
-    version = "2020.3"
+    version = "LATEST-EAP-SNAPSHOT"
}
+
+ tasks.patchPluginXml {
+     sinceBuild("203")
+ }
```

I'm really no Jetbrains Plugin expert, but from my understanding this is how it should be working.

cgrushko:
ok; let's test it tomorrow (it's night time in my timezone).

jef:
I recently updated one of my [plugins](https://github.com/jef/forest-night-jetbrains/commit/8996da8298d27eed2d92adc7fcb1a48d14386cab) and it got approved and seems to be working. That's somewhat reassuring.

I'll test in the meantime and see if you get similar results!

jef:
If I remove

```diff
- tasks {
-     patchPluginXml {
-         sinceBuild("203")
-     }
- }
```

altogether, then it also works... I think the real issue here is

```diff
intellij {
-    version = "2020.3"
+    version = "LATEST-EAP-SNAPSHOT"
}
```

That being said, Not sure what benefit `sinceBuild` gives other than their [shotty doc](https://github.com/JetBrains/gradle-intellij-plugin#patching-dsl). Relates to their [building properties](https://github.com/JetBrains/gradle-intellij-plugin#building-properties), which also don't give great insight (imo).

Anyway, take it as you see it. We may only need to update `version`.

cgrushko:
mm, so we're increasing the minimum to EAP, and the old versions keep living in JetBrain's plugin repo, and older versions of IJ fetch the older versions of the client?
Alright, sounds reasonable - let's give it a try :) 
I'll pull it into our internal repo and after review it'll be pushed back with your username as authoring it. 
Thanks!

jef:
> the old versions keep living in JetBrain's plugin repo, and older versions of IJ fetch the older versions of the client?

Sounds like a reasonable assumption.

> Alright, sounds reasonable - let's give it a try :)
> I'll pull it into our internal repo and after review it'll be pushed back with your username as authoring it.

Great! Let me know if anything comes up.

jef:
All is well. I think I understand the `sinceBuild` now. Probably doesn't become available until 203*. Not sure if it's needed really. I would be interested in the side effects of removing.

PR #156, opened 2020-12-26
Add support to all versions above 2020.1.4, including

JavierSegoviaCordoba:
Now Android Studio 4.1.1 is compatible too which is the latest stable version. It fixes https://github.com/facebookincubator/ktfmt/issues/61

To get it working with Android Studio 4.1.1, I need to downgrade from Java 11 to Java 8.

The plugin can be installed in AS 4.1.1, can be enabled, and change its settings, but it doesn't work when you try to reformat (it does nothing). Meanwhile, the same plugin installed in the latest AS Canary or in IntelliJ works.

### DIFF
--- a/ktfmt_idea_plugin/build.gradle.kts
+++ b/ktfmt_idea_plugin/build.gradle.kts
@@ -30,8 +30,8 @@ repositories {
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_11
-    targetCompatibility = JavaVersion.VERSION_11
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
 }
 
 dependencies {
@@ -39,13 +39,18 @@ dependencies {
     implementation("com.google.googlejavaformat", "google-java-format", "1.8")
 }
 
-// See https://github.com/JetBrains/gradle-intellij-plugin/
+/**
+ * Minimum compatible version.
+ *
+ * `intellij.version` should match `sinceBuild(value)`
+ */
 intellij {
-    version = "LATEST-EAP-SNAPSHOT"
+    version = "2020.1.4"
 }
 
 tasks {
     patchPluginXml {
-        sinceBuild("203")
+        sinceBuild("201")
+        untilBuild("")
     }
 }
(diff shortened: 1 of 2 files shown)


JavierSegoviaCordoba:
The plugin can be installed in AS 4.1.1, can be enabled, and change its settings, but it doesn't work when you try to reformat (it does nothing). Meanwhile, the same plugin installed the latest AS canary, and IntelliJ works.

cgrushko:
@JavierSegoviaCordoba , thanks for the PR!
Questions - 
1. google-java-format doesn't work with Java 8, so I'd expect downgrading to 8 to break everything. Can you elaborate on the downgrading part?
2. This PR makes the plugin compatible with 2020.1.4, but it doesn't make it work? can you elaborate on which versions will work?

JavierSegoviaCordoba:
1. I am downgrading it because I can't install it with Java 11 in AS 4.1.1.
2. I am using it for days without problems in Android Studio 4.2 Beta. IntelliJ 2020.3.1 RC works with it too but I didn't try it a lot. I installed it in 4.1.1 but when I press the reformat hotkey it says there is nothing to reformat, but it can be enabled at least.
 
I have no idea why it is not working in AS 4.1.1, but I kept the PR open so maybe you can try to check or improve it if you know how (I am not a plugin expert, to be honest).

JavierSegoviaCordoba:
Maybe it is not broken on newer version even with Java 8 because the newer version are using Java 11+ and AS 4.1.1 is using Java 8?

Just speculating 🤔

cgrushko:
@JavierSegoviaCordoba , I ended up with https://github.com/facebookincubator/ktfmt/pull/159, which is similar to this PR but adds documentation and notes based on my testing today. I wasn't able to make it work with v193, but Java 8 doesn't seem to be needed while the extra method there is. So, I'm going to go ahead with https://github.com/facebookincubator/ktfmt/pull/159. Thanks a lot for contributing :)

JavierSegoviaCordoba:
@cgrushko I am checking and ktfmt is not compatible with 2021 EAP. Maybe the issue I mentioned is wrong and should be used `version = "LATEST-EAP-SNAPSHOT"`?

cgrushko:
@JavierSegoviaCordoba that's expected - see the description of versions in https://github.com/facebookincubator/ktfmt/blob/e7803570736cf1017f80ef8dab66479d01bf18fb/ktfmt_idea_plugin/DEVELOPING.md . 

I used a concrete version to make the build predictable (or, repeatable). Using `LATEST-EAP-SNAPSHOT` means it will start breaking without warning.

If you've tested the plugin with version 211 (2021.1) then let's bump the version and I'll push a new version.

JavierSegoviaCordoba:
I was trying this afternoon, but `gradlew build` throws that a test is failing

PR #157, opened 2020-12-29
Use ktfmt 0.19 in IDEA plugin

sgrimm:
It looks like `bump_version.sh` will update the plugin from version 0.20 on, but this needs to be done manually for 0.19.

### DIFF
--- a/ktfmt_idea_plugin/build.gradle.kts
+++ b/ktfmt_idea_plugin/build.gradle.kts
@@ -19,7 +19,7 @@ plugins {
     java
 }
 
-val ktfmtVersion = "0.18"
+val ktfmtVersion = "0.19"
 
 group = "com.facebook"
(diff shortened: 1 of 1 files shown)


sgrimm:
I assume releasing the new plugin to the plugin repository is a manual process someone at Facebook has to do, but happy to update this PR if it requires additional changes to the code.

cgrushko:
Thanks! yeah, if I'm being honest I can do it locally without landing it at all.
Let me quickly release 0.19 and then let's take a look at https://github.com/facebookincubator/ktfmt/pull/156 .

cgrushko:
I ended up writing https://github.com/facebookincubator/ktfmt/pull/159 which fixes several problems in one go. I also submitted a new plugin version based on 0.19 to JetBrains, so it should hopefully show up in a couple of days. Thanks for contributing!

PR #168, opened 2021-02-01
Make IntelliJ plugin works with 2021-EAP-SNAPSHOT

JavierSegoviaCordoba:


ktfmt_idea_plugin/build.gradle.kts
```diff
@@ -42,7 +42,7 @@ dependencies {
 // See https://github.com/JetBrains/gradle-intellij-plugin/
 intellij {
     // Version with which to build (and run; unless alternativeIdePath is specified)
-    version = "2020.3"
+    version = "211-EAP-SNAPSHOT"
```
cgrushko, line 45: shouldn't this be "211"? do we expect the API to change much until it's released? if so, perhaps we shouldn't land it here yet, and instead install from disk.

JavierSegoviaCordoba:
I think it can be 211, if bumping from 203 to 211 has no changes probably it will work with the future 211 release.

I am going to try with 211 and I will update it

JavierSegoviaCordoba:
It doesn't allow use 211 (203 doesn't work too). They are too generic. 2021.1 doesn't exist too.

After checking the [docs](https://github.com/JetBrains/gradle-intellij-plugin/#intellij-platform-properties):

Acceptable Values:
- version : `2017.2.5` or `IC-2017.2.5`
- build: `172.4343` or `IU-172.4343`
- `LATEST-EAP-SNAPSHOT`

So the only thing we can do is wait for the release.

I have a doubt, a lot of plugins I use didn't update and they are compatible with 211 builds. They updated some time ago to get it working? For example, detekt IDE plugin was updated 2020/09... and its config is:

```
configure<IntelliJPluginExtension> {
    pluginName = "Detekt IntelliJ Plugin"
    version = "2019.3"
    updateSinceUntilBuild = false
    setPlugins("IntelliLang", "Kotlin")
}
```

cgrushko:
Yeah.. that's weird. I wonder if we can extract the xml from another plugin and reverse engineer it.

JavierSegoviaCordoba:
The detekt plugin is open source: https://github.com/detekt/detekt-intellij-plugin

This xml? https://github.com/detekt/detekt-intellij-plugin/blob/master/src/main/resources/META-INF/plugin.xml

JavierSegoviaCordoba:
@cgrushko what do you think about a GitHub Action which generates the Gradle plugin with the last snapshot so everyone can download the zip without building the project? We can just set it as `workflow_dispatcher` and when a new version appears, we just have to click a button and the GitHub package will be there.

cgrushko:
@JavierSegoviaCordoba  sounds great :) I tried writing one to automate our release but couldn't find a way to trigger a workflow from a button on the page. Is this something you can write a PR for, or does it have to be tested with facebookincubator's autho?

JavierSegoviaCordoba:
I can create a PR for you, but you will have to test it, in a few hours I will do :)

PR #174, opened 2021-02-17
GoogleStyle: Use ZERO indent in IF expressions for simple binary expressions

bethcutler:


cgrushko (changes requested):
The PR doesn't work for these cases:

```
  if (expressions1
      .expression2
  ) {
    bar()
  }
```
…
```
  if (!(expressions1 &&
      expression2 &&
      expression3)
  ) {
    bar()
  }
```
…

I'm trying to balance simplicity with getting things done with how important this styling is, and in this case I'd rather we keep looking for a better solution....

JavierSegoviaCordoba:
Google Style is the default one? If not, is it documented? And is it available via Spotless? I am not sure because checking the default blocking indent is two, and Google Style is 4.

cgrushko:
“Google style” is... not the Android / JetBrains style :/
It isn’t the default, and isn’t available through Spotless.
We’re not 100% sure it works properly yet, so we didn’t publicize it.
We didn’t plan on deriving a JetBrains style from it, but it’s a very
reasonable thing to do at some point.

JavierSegoviaCordoba:
I deleted my comment because I was checking and old commit, but it seems I haven't wrong hahah

Yeah, I think should be great to have a JetBrains style.

PR #182, opened 2021-02-26
fix: version compatibility

jef:
### Description

Uses the default (`LATEST-EAP-SNAPSHOT`) for `version`, since we probably want to build on latest.

Also changes `updateSinceUntilBuild` to `false` from it's default `true`. I believe this is good to address issues where the plugin isn't available for EAP and other JetBrains IDEs.

See https://github.com/JetBrains/gradle-intellij-plugin#building-properties.

Similar to Prettier's plugin: https://github.com/JetBrains/intellij-plugins/blob/master/prettierJS/build.gradle.kts

---

Here is a demo of one of my plugins ([release pull request](https://github.com/jef/forest-night-jetbrains/pull/24)) for verification:

image

vs not having it:

image

### DIFF
--- a/.github/workflows/plugin.yml
+++ b/.github/workflows/plugin.yml
@@ -1,4 +1,4 @@
-name: Build ktfmt JetBrains plugin
+name: Build JetBrains plugin
 
 on:
   push:
--- a/.github/workflows/plugin.yml
+++ b/.github/workflows/plugin.yml
@@ -31,4 +31,4 @@ jobs:
     - name: Verify plugin
       run: |
         cd ktfmt_idea_plugin
-        ./ktfmt_idea_plugin/gradlew verifyPlugin
+        ./gradlew verifyPlugin
--- a/.github/workflows/plugin.yml
+++ b/.github/workflows/plugin.yml
@@ -1,0 +1,34 @@
+name: Build ktfmt JetBrains plugin
+
+on:
+  push:
+    paths-ignore:
+      - '**.md'
+  pull_request:
+    paths-ignore:
+      - '**.md'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # test against latest update of each major Java version, as well as specific updates of LTS versions:
+        java: [ 11, 13 ]
+    name: Build plugin on Java ${{ matrix.java }}
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        submodules: recursive
+    - name: Set up JDK ${{ matrix.java }}
+      uses: actions/setup-java@v1
+      with:
+        java-version: ${{ matrix.java }}
+    - name: Build plugin
+      run: |
+        cd ktfmt_idea_plugin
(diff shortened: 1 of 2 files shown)


jef:
Feel free to change the `sinceBuild` to like `193` or something.

Source: https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html#intellij-platform-based-products-of-pre-20162-ide-versions

I forget when this was first released, but I believe it was `193`.

cgrushko:
@JavierSegoviaCordoba - you tried something similar, right?

JavierSegoviaCordoba:
Yeah, the problem I got is the plugin doesn't compile with LATEST-EAP-SNAPSHOT, I have to update some methods that have changed in order to use the plugin in 2021.1

JavierSegoviaCordoba:
@jef can you verify you can build and install the plugin?

jef:
Sorry, I was under the assumption that the CI did that, but I'm realizing now that it is for the rest of the project.

That being said, I couldn't get it to run without getting a few lines of these:

```
CodeStyleManagerDecorator.java:92: error: name clash: reformatText(PsiFile,Collection<TextRange>) in CodeStyleManagerDecorator and reformatText(PsiFile,Collection<? extends TextRange>) in CodeStyleManager have the same erasure, yet neither overrides the other
  public void reformatText(PsiFile file, Collection<TextRange> ranges)
```

I believe the same thing that @JavierSegoviaCordoba ran into.

---

Building on 2020.3 worked though and the key part here is `updateSinceUntilBuild`.

I thought it would be good to add workflows for building the plugin as well. Not necessary, but could be useful if we decide to pull that in. Let me know otherwise.

JavierSegoviaCordoba:
I think the only way to allow users to get the new version, is merging the PR I created which generate and upload the plugin as a zip in GitHub Actions.

Steps:

1. Create a permanent branch called, for example, latest-eap-snapshot. 

2. Update it continuously so latest eap compiles.

3. Launch the workflow from GitHub GUI to generate a zip which can be downloaded.

jef:
> I think the only way to allow users to get the new version, is merging the PR I created which generate and upload the plugin as a zip in GitHub Actions.

I don't believe that's entirely true. I have a workflow that pushes to the marketplace directly

https://github.com/jef/forest-night-jetbrains/blob/main/.github/workflows/release.yaml

---

That being said, it can still be done manually and this PR ensures that we also don't need to publish a version without any code change.

JavierSegoviaCordoba:
But ktfmt should not be published with latest eap until it is stable, sample:

ktfmt 0.20 is published to marketplace with 2021.1 compatibility.

ktfmt 0.21 is released, you have to rollback the 2021.1 changes to 2020.3 and publish it because if you publish the plugin with 0.21, it will not work on 2020.3

Maybe the plugin from the latest eap branch can be published to a different channel to avoid downloading the plugin as zip from actions, I can transform the workflow I created to do this.

What do you think @cgrushko?

PR #183, opened 2021-02-26
Add Google and Kotlinlang support to IDEA plugin

JavierSegoviaCordoba:
Now the drop-down has 4 items:

Default, Dropbox, Google and Kotlinlang. Should I delete Dropbox?

### DIFF
--- a/ktfmt_idea_plugin/src/main/java/com/facebook/ktfmt/intellij/FormatterUtil.java
+++ b/ktfmt_idea_plugin/src/main/java/com/facebook/ktfmt/intellij/FormatterUtil.java
@@ -16,15 +16,13 @@
 
 package com.facebook.ktfmt.intellij;
 
-import static com.facebook.ktfmt.FormatterKt.DROPBOX_FORMAT;
-
 import com.facebook.ktfmt.FormatterKt;
-import com.facebook.ktfmt.FormattingOptions;
 import com.facebook.ktfmt.ParseError;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.googlejavaformat.java.FormatterException;
 import com.intellij.openapi.util.TextRange;
+
 import java.util.Map;
 
 final class FormatterUtil {
@@ -36,16 +34,18 @@ final class FormatterUtil {
    *
    * @return formatted code
    */
-  static Map<TextRange, String> getReplacements(boolean isDropboxStyle, String code) {
+  static Map<TextRange, String> getReplacements(UiFormatterStyle uiFormatterStyle, String code) {
     try {
-      return ImmutableMap.of(TextRange.allOf(code), formatCode(isDropboxStyle, code));
+      return ImmutableMap.of(TextRange.allOf(code), formatCode(uiFormatterStyle, code));
     } catch (FormatterException | ParseError e) {
       return ImmutableMap.of();
     }
   }
 
   @VisibleForTesting
-  static String formatCode(boolean isDropboxStyle, String code) throws FormatterException {
-    return FormatterKt.format(isDropboxStyle ? DROPBOX_FORMAT : new FormattingOptions(), code);
+  static String formatCode(UiFormatterStyle uiFormatterStyle, String code)
+      throws FormatterException {
+
+    return FormatterKt.format(uiFormatterStyle.getFormattingOptions(), code);
   }
 }
--- a/ktfmt_idea_plugin/src/main/java/com/facebook/ktfmt/intellij/KtfmtConfigurable.java
+++ b/ktfmt_idea_plugin/src/main/java/com/facebook/ktfmt/intellij/KtfmtConfigurable.java
@@ -25,15 +25,16 @@ import com.intellij.openapi.ui.ComboBox;
 import com.intellij.uiDesigner.core.GridConstraints;
 import com.intellij.uiDesigner.core.GridLayoutManager;
 import com.intellij.uiDesigner.core.Spacer;
-import java.awt.Insets;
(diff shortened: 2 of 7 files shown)


ktfmt_idea_plugin/src/main/java/com/facebook/ktfmt/intellij/FormatterUtil.java
```diff
@@ -16,36 +16,56 @@
...
       return ImmutableMap.of();
     }
   }
 
   @VisibleForTesting
-  static String formatCode(boolean isDropboxStyle, String code) throws FormatterException {
-    return FormatterKt.format(isDropboxStyle ? DROPBOX_FORMAT : new FormattingOptions(), code);
+  static String formatCode(UiFormatterStyle uiFormatterStyle, String code) throws FormatterException {
+
+    FormattingOptions formattingOptions;
```
cgrushko, line 52: `final FormattingOptions formattingOptions;`

will make the compiler verify it's set exactly once, which makes it safer.

But - see my comment in KtfmtSettings.java before working on this file.

ktfmt_idea_plugin/src/main/java/com/facebook/ktfmt/intellij/KtfmtConfigurable.java
```diff
@@ -25,16 +25,13 @@
...
-import javax.swing.JCheckBox;
-import javax.swing.JComboBox;
-import javax.swing.JComponent;
-import javax.swing.JLabel;
-import javax.swing.JPanel;
 import org.jetbrains.annotations.Nls;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import javax.swing.*;
```
cgrushko, line 32: replace .* with individual imports

ktfmt_idea_plugin/src/main/java/com/facebook/ktfmt/intellij/UiFormatterStyle.java
```diff
@@ -17,33 +17,28 @@
...
-  DROPBOX("Dropbox style", true);
+  DEFAULT("Default"),
+  DROPBOX("Dropbox"),
+  GOOGLE("Google"),
+  KOTLINLANG("Kotlinlang");
 
   private final String description;
-  private final boolean isDropboxStyle;
 
-  UiFormatterStyle(String description, boolean isDropboxStyle) {
```
cgrushko, line 30: instead of `boolean isDropboxStyle`, use `FormattingOptions formattingOptions`.
Then, initialize like that -
```
DEFAULT("Default", new FormattingOptions()),
DROPBOX("Dropbox", com.facebook.ktfmt.DROPBOX_FORMAT),
GOOGLE("Google (internal)", GOOGLE_FORMAT),
KOTLINLANG("kotlinlang.org", KOTLINLANG_FORMAT),

then, you don't need the switch in FormatterUtil.java, you can directly use the field.

```

ktfmt_idea_plugin/src/main/java/com/facebook/ktfmt/intellij/UiFormatterStyle.java
```diff
@@ -17,33 +17,28 @@
...
   public String toString() {
     return description;
   }
 
-  public boolean convert() {
-    return isDropboxStyle;
-  }
-
-  static UiFormatterStyle convert(boolean isDropboxStyle) {
+  static UiFormatterStyle convert(UiFormatterStyle uiFormatterStyle) {
```
cgrushko, line 39: we don't need this method anymore - delete it entirely

cgrushko (changes requested):
Thank you very much for this contribution!
I left some comments, I hope they make sense.
Thanks again!

ktfmt_idea_plugin/src/main/java/com/facebook/ktfmt/intellij/FormatterUtil.java
JavierSegoviaCordoba: That is very funny, I was trying to use `switch` like it was an enum in Kotlin until I remembered I can't.

ktfmt_idea_plugin/src/main/java/com/facebook/ktfmt/intellij/FormatterUtil.java
```diff
@@ -16,36 +16,56 @@
...
 import com.facebook.ktfmt.FormattingOptions;
 import com.facebook.ktfmt.ParseError;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.googlejavaformat.java.FormatterException;
 import com.intellij.openapi.util.TextRange;
+
 import java.util.Map;
 
+import static com.facebook.ktfmt.FormatterKt.*;
```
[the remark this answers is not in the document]
JavierSegoviaCordoba: I am checking that the order of some imports in multiple files has changed. I was using the Google plugin plus reformat code + optimize imports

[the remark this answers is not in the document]
cgrushko: Unfortunately the GJF plugin doesn't sort imports (unclear why). In general, to preserve blame and history it's best not to change anything unrelated to your change.
I'll sort them after importing to our internal repo, our linters will complain anyway.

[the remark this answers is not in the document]
JavierSegoviaCordoba: Spotless does? It has Google format for java.

PR #184, opened 2021-02-28
Add Spotless (GoogleJavaFormat) to IDEA plugin

JavierSegoviaCordoba:
To ensure the IDEA plugin code is formatted correctly, Spotless will check it every time a push or PR.

I tweaked the plugin workflow created by @jef to just use `./gradlew build` which depends on the task `spotlessJavaCheck`, so it will check if the plugin code compiles, tests pass, and if the code is correctly formatted.

To automatically format the code, run `./gradlew spotlessJavaApply`.

### DIFF
--- a/.github/workflows/plugin.yml
+++ b/.github/workflows/plugin.yml
@@ -1,0 +1,28 @@
+on:
+  push:
+    paths-ignore:
+      - '**.md'
+  pull_request:
+    paths-ignore:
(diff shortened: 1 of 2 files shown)


cgrushko:
Beautiful :) 
(after importing, we'll need to wait for someone in the US to review, so probably Monday night Spanish time)

PR #187, opened 2021-03-05
Add an `.editorconfig` file per style

JavierSegoviaCordoba:
I would ask for help from the community so we can fix all styles by updating this PR.

Currently, all styles are using the same base file provided by @sgrimm, the only changes among them are the properties `indent_size` and `ij_continuation_indent_size` (and a few properties related to avoiding wildcard imports), but I think there are more properties that should be updated/added/removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,8 @@ shortcut.
 To configure IntelliJ to approximate ktfmt's formatting rules during code editing,
 you can edit your project's
 [`.editorconfig` file](https://www.jetbrains.com/help/idea/configuring-code-style.html#editorconfig)
-to include the Kotlin section from
-[`docs/editorconfig-section.txt`](docs/editorconfig-section.txt).
+to include the Kotlin section from one of the files inside
+[`docs/editorconfig`](docs/editorconfig).
 Not all of ktfmt's rules can be represented as IntelliJ editor settings, so you will still
 need to run ktfmt. Alternately, that file can be used as a reference to manually change
 the project's code style settings.
--- a/docs/editorconfig/.editorconfig-default
+++ b/docs/editorconfig/.editorconfig-default
@@ -4,7 +4,7 @@
 #
 # It includes editor-specific config options for IntelliJ IDEA.
 #
-# To use Dropbox style, uncomment the lines at the bottom of the section.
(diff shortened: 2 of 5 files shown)


cgrushko (commented):
This looks great, but since I don't use `.editorconfig` files it's hard for me to give meaningful feedback.

@sgrimm , do you want to take a look?

Given that this is something that's hard for the maintainers to give feedback on, would it make sense to create a separate repository for these files, and point to them from ktfmt's README?

JavierSegoviaCordoba:
I personally use them as fallback so if someone in the team doesn't install the IntelliJ plugin, can at least get a similar formatting before checking the code with the Gradle plugin.

I think IntelliJ IDEA lets to export the .editorconfig, but I don't know all the specific configs of all styles.

Should the README point to the conventions docs of each style to know more specific differences and not only if it is 2/4, 2/2, 4/8, etc?

sgrimm:
Having separate per-style files seems like the right approach to me. I have no opinion to offer about whether the settings for the other styles are correct, though, since I don't use those styles. As I said in my initial PR for this, I'm not even all that confident my original file was exactly correct, just closer to ktfmt default style than the default IntelliJ settings are.

My use case is exactly the same as @JavierSegoviaCordoba says: a fallback to reduce the number of complaints from the code style checker if a team member doesn't install the plugin.

cgrushko (changes requested):
@JavierSegoviaCordoba , should we rename or delete editorconfig-section.txt?
should we add a similar comment block to the one in editorconfig-section.txt to the other files?

JavierSegoviaCordoba:
I think we should delete it, I think it is easy to just open/download the correct file and copy/rename it in the project.

The problem is how we can get similar behavior in all those files because nobody has the knowledge to know how to complete them.

Maybe we can just leave this PR open until we get help for each file or we can just merge it and if somebody finds a problem with any .editorconfig file, a new PR can be open to fix it.

cgrushko:
@JavierSegoviaCordoba let's do this for now - 
1. delete the old file
2. add some comments to the new files, probably copy it from the old file. Make it clear that it's WIP and we're happy to get PRs.
what do you think?

PR #214, opened 2021-04-22
Update ktfmt to fix odd formatting with DSLs.

steventrouble:
Update KotlinInputAstVisitorBase.kt's binary expression handling to treat trailing lambdas as separate from the rest of the binary expression. While this should clean up some issues with binary expressions, the main intended side-effect was to fix DSLs:

https://github.com/facebookincubator/ktfmt/issues/206
    
Before, it would format DSLs like this:

```
a =
  a {
    b =
      b {
        c = c
      }
  }
```

After, it should format them like this

```
a = a {
  b = b {
    c = c
  }
}
```


core/src/test/java/com/facebook/ktfmt/FormatterKtTest.kt
```diff
@@ -350,6 +350,18 @@ class FormatterKtTest {
         |""".trimMargin(),
           deduceMaxWidth = true)
 
+  @Test
+  fun `binary operators dont break when the last one is a lambda`() =
+      assertFormatted(
+          """
+      |----------------------
+      |foo =
+      |    foo + bar + dsl {
```
cgrushko, line 359: can you add a test case where there's a line break in the middle of the binary expression?

cgrushko (commented):
Thanks! this is awesome! and looks great.

1. I'll run the modified code on our repo to see if it does anything weird. I'll report back.
2. I commented about a test case that would show how line breaks affect this change. If you can think of additional ones it'll increase confidence in the change.

core/src/test/java/com/facebook/ktfmt/FormatterKtTest.kt
steventrouble: Good idea, on it.

(Sorry for the delay here, just got back from vacation.)

cgrushko:
@steventrouble , thanks again for the PR and really sorry it took me forever to look at it.
I ran the modified ktfmt on our codebase, and it drastically improves a large number of lambdas - that's awesome :)

I ran into two weird results of the PR. I'll write a sanitized version of them ASAP and post here. 
Meanwhile, I'll describe -
1. `a * b + c * d + e * f` are broken into
a * b +
c * d + 
e * f
This is inside an `if`, though there is a lambda later in the expression.

2. `foo.bar ?: 0 > 0` is broken into
```
foo.bar
  ?: 0 > 0
```

Can you take a look into why this might happen?

cgrushko:
Please hold on regarding my previous comment - I might have looked at the wrong diff, this might be an existing problem.

Issue #235, opened 2021-05-17
Unpaired ``` not gracefully handled


bethcutler:
Original code; triple quote is unpaired accidentally
```kotlin
/**
 * ```kotlin
 * class foo {
 * }
 */
```

After formatting; an even # of triple quotes are added, which does nothing to solve the problem.  Every time formatter is run, more triple quotes are added.
```kotlin
/**
 * ```kotlin
 * ```
 * ```
 * class foo { }
 * ```
 * ```
 */
```

bethcutler:
Here's a similar example, in which the triple quotes are *not* unpaired:

Before:
```kotlin
  /**
   * Title ```block```.
   * More text.
   */
```

After formatting once:
```kotlin
  /**
   * Title ```block```.
   * ```
   * ```
   * More text.
   * ```
   * ```
   */
```

Each time the file is formatted, 2 more lines of triple quotes are added before and after.

bethcutler:
I started looking into this, and it looks like the problem stems from kdoc lexer tagging whatever token follows "```block```." as a CODE_BLOCK_TEXT token, which I would assume is a bug.

Not sure if we can fix kdoc lexer.  But also not sure what an appropriate workaround would be.

bethcutler:
Update: the first issue is indeed fixed, but the second case is still not ideal.

Original:
```kotlin
/**
 * Title ```block```.
 * More text.
 */
```

Formatted:
```kotlin
/**
 * Title
 * ```block```.
 * More text.
 * ```
 */
```

Adding a new triple quote seems wrong in this case.

PR #238, opened 2021-05-27
Fix for formatting on tagged code samples in KDoc.

bethcutler:
Fix for #235


cgrushko:
Looks great! I imported the PR and it should land next US week when someone additional has a chance to take a look.
The two commits in this PR will unfortunately be squashed into a single commit once (as a result of Facebook's equivalent of Copybara). My experience with Git is limited, but one way to avoid this is to put each fix in its own branch.

PR #244, opened 2021-06-23
Example of calling `visitCallElement` instead of using custom logic.

steventrouble:
This is an example of calling `visitCallElement` in `processLambdaOrScopingFunction`. It allows the process function to handle lambdas with arguments or type arguments.

Also as an example, this also modifies the `lambdaOrScopingFunction` to return `true` for lambdas with arguments or type arguments.

While this passes unit tests, **I have no real-life code to test this against** because I'm using a different box. My dev box is currently having some auth issues working with github and it's taking a while to resolve them.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/KotlinInputAstVisitorBase.kt
+++ b/core/src/main/java/com/facebook/ktfmt/KotlinInputAstVisitorBase.kt
@@ -1225,9 +1225,9 @@ open class KotlinInputAstVisitorBase(
    * 1. '... = { ... }'
    * 2. '... = Runnable { ... }'
    * 3. '... = scope { ... }' '... = apply { ... }'
+   * 4. '... = foo() { ... }'
    *
    * but not:
-   * 1. '... = foo() { ... }'
    * 2. '... = Runnable @Annotation { ... }'
    */
   private fun lambdaOrScopingFunction(initializer: PsiElement?): Boolean {
@@ -1235,9 +1235,7 @@ open class KotlinInputAstVisitorBase(
       return true
     }
     if (initializer is KtCallExpression &&
-        initializer.valueArgumentList?.leftParenthesis == null &&
         initializer.lambdaArguments.isNotEmpty() &&
-        initializer.typeArgumentList?.arguments.isNullOrEmpty() &&
         initializer.lambdaArguments.first().getArgumentExpression() is KtLambdaExpression) {
       return true
     }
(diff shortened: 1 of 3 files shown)


msfjarvis:
I tested it against [Android-Password-Store](https://github.com/Android-Password-Store/Android-Password-Store) where I noticed the regression and while the crash is fixed, I find the formatting to be subpar in multiple places. While it works well [here](https://github.com/msfjarvis/Android-Password-Store/commit/dcf8ad4df694#diff-976b0f146a4f6bca184021a2b27fdd5f1d3fe88f12ae686e7c91313ec49bff1d) in an assignment context, it performs poorly in a longer line width variant in the [very next file](https://github.com/msfjarvis/Android-Password-Store/commit/dcf8ad4df694#diff-a8def017660e4088e43f5a75e5d2dc657490d646cce3d383ab23e416a4bbf7ba). The problems are more apparent when viewing a [split diff](https://github.com/msfjarvis/Android-Password-Store/commit/dcf8ad4df694390f5a22753d74472034e1dddc59?branch=dcf8ad4df694390f5a22753d74472034e1dddc59&diff=split).

steventrouble:
Thanks for the diffs! The fact that the function call is on the same indent as the params is definitely an issue. I'll investigate.

---

Btw, I think some of the issues are not with this particular fix. Specifically, the following is a [known issue with trailing lambdas](https://github.com/facebookincubator/ktfmt/blob/cd09a6c887391301c8f3a80f5ba1ae2ae615cbe7/core/src/main/java/com/facebook/ktfmt/KotlinInputAstVisitorBase.kt#L1251) from before the change:

```
  // Known issue:
  val foo =
      bar {  // This is double indented
    result  // This is only single indented
  }
```

The only reason it didn't show up on those lines before was because we weren't handling trailing assignment lambas in the same way as declaration lambdas. Now that they're using a common code path (as IMO they should be), the bug is more visible. In other words, the bug was always there, but was being hidden by another bug.

We're still thinking about how to fix the trailing lambda, but here's a quick doc on why it's taking so long and is such a hard issue to fix:

https://docs.google.com/presentation/d/1EClwBAz-VNOGva3da1AclPUnptBO6rETVcAhy-n6Ic4/view?usp=sharing

JavierSegoviaCordoba:
@cgrushko this has been fixed with 0.33, no?

cgrushko:
Sorry, I lost track of what this PR is fixing :\

hick209:
Closing as as it's too old at this point.

@steventrouble feel free to reopen it if you believe this is still an issue

PR #263, opened 2021-12-14
Support `expect enum class` without a body.

mpetrov:
Currently the parser breaks when encountering these in mutliplatform
projects.

Note that `kotlinc` also supports empty enum definitions.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -1299,7 +1299,9 @@ class KotlinInputAstVisitor(
       }
       val body = classOrObject.body
       if (classOrObject.hasModifier(KtTokens.ENUM_KEYWORD)) {
-        visitEnumBody(classOrObject as KtClass)
+        if (body != null || !classOrObject.hasModifier(KtTokens.EXPECT_KEYWORD)) {
+          visitEnumBody(classOrObject as KtClass)
+        }
       } else if (body != null) {
         visitBlockBody(body, true)
       }
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -1299,9 +1299,7 @@ class KotlinInputAstVisitor(
       }
       val body = classOrObject.body
       if (classOrObject.hasModifier(KtTokens.ENUM_KEYWORD)) {
-        if (body != null) {
-          visitEnumBody(classOrObject as KtClass)
-        }
+        visitEnumBody(classOrObject as KtClass)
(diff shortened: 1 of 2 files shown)


hick209 (approved):
Thanks for helping improve ktfmt

mpetrov:
I don't have permissions to run the `Build and Test`  check or merge, so let me know if there's any more action needed on my end.

mpetrov:
Updated due to conflict with `main`. Sorry but will need approval again to rerun the workflows since I'm a first time contributor, and I'd also need someone authorized to merge to actually merge the request afterwards.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -1299,7 +1299,9 @@ class KotlinInputAstVisitor(
       }
       val body = classOrObject.body
       if (classOrObject.hasModifier(KtTokens.ENUM_KEYWORD)) {
-        visitEnumBody(classOrObject as KtClass)
+        if (body != null) {
+          visitEnumBody(classOrObject as KtClass)
```
cgrushko, line 1303: does it make sense checking if `body` is null inside `visitEnumBody`? that method is only called from here right now, but I think it makes it easier to read the code overall.

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -4509,4 +4509,10 @@ class FormatterTest {
       |  y
       |}
       |""".trimMargin())
+
+  @Test
+  fun `expect enum class`() =
```
cgrushko, line 4514: can you move this alongside the other `enum` tests? around line 3390

cgrushko (changes requested):

cgrushko:
@mpetrov , thank you very much for this contribution! I left a couple of comments.

Re merging - since ktfmt's source of truth is Meta's internal repo, we won't merge this PR directly. Instead, we'll land the diff that was created above (mentioned in https://github.com/facebookincubator/ktfmt/pull/263#issuecomment-993979043) and then it'll be exported to the repo on GitHub. You'll be credited as the author.

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
mpetrov: Good call, done.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
mpetrov: SGTM, done.

mpetrov:
I assumed github was the source of truth, thanks for the context @cgrushko! Made the suggested changes.

cgrushko:
Started landing in the source-of-truth repo, this PR should auto-close when it lands

PR #264, opened 2021-12-20
Fix main class in pom file.

mpetrov:
Building with `mvn install` currently results in a jar with a missing main class, as `main` was moved to core/src/main/java/com/facebook/ktfmt/cli/Main.kt


hick209:
Thanks for the fix, @mpetrov!

This also points out something we need to include to our testing to make sure it we don't break this going forward. @davidtorosyan is working on it.

hick209 (approved):

hick209 (approved):

PR #266, opened 2021-12-28
Update kotlin version to 1.6.10

mpetrov:
Running `mvn install` on macOS with JDK15 results in illegal access
exceptions. This appears to be a known issue that was fixed in 1.5.30:
https://youtrack.jetbrains.com/issue/KT-45689

It would be a nice quality of life improvement to bump the kotlin
version if this doesn't conflict with any internal flows.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -16,7 +16,7 @@
 
     <properties>
         <dokka.version>0.10.1</dokka.version>
-        <kotlin.version>1.5.0</kotlin.version>
+        <kotlin.version>1.5.32</kotlin.version>
         <kotlin.compiler.incremental>true</kotlin.compiler.incremental>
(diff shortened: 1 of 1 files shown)


cgrushko:
@mpetrov thanks! Maybe it's worth upgrading to 1.6?

mpetrov:
Sure, no issues bumping to 1.6.10 that I'm aware of.

@cgrushko PTAL

cgrushko:
Thanks! Waiting for the importer to pick this up. We don't have good coverage for the binary itself, any chance I can ask you to build and run the binary for sanity check?

mpetrov:
Yup, got `No changes` on all files after running:

```
mvn clean
mvn install
find core/src/ | egrep ".kt$" | xargs java -jar core/target/ktfmt-0.31-SNAPSHOT-jar-with-dependencies.jar
```

This is on macOS with `java 15.0.2 2021-01-19`.

PR #275, opened 2022-02-08
Account for semicolons on top-level statements.

nreid260:
Before this, such semicolons caused a crash because OpBuilder noticed they were skipped in the token stream.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -2293,6 +2293,7 @@ class KotlinInputAstVisitor(
         builder.blankLineWanted(OpsBuilder.BlankLineWanted.YES)
       }
       visit(child)
+      builder.guessToken(";")
       lastChildHadBlankLineBefore = childGetsBlankLineBefore
       first = false
     }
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -2296,6 +2296,7 @@ class KotlinInputAstVisitor(
       lastChildHadBlankLineBefore = childGetsBlankLineBefore
       first = false
     }
+    builder.guessToken(";")
```
cgrushko, line 2299: 1. I would have expected this to be after `visit(child)`, no?
2. We had a nasty bug in the past where guessToken would swallow a token from a completely unrelated node in the AST, so I'm weary of it. I don't remember the specifics unfortunately. Any chance you can poke around the expression object and maybe find a boolean indicating whether there's a trailing semicolon?

nreid260: Yes, fixed. I'm cut/paste error from my main workspace.

nreid260: I don't see any such boolean on `KtExpression`. Perhaps we could check for it in every `visit` method, but that seems pretty error prone.

I'm also not sure how this could manifest such a bug. These semicolons are all immediately after a statement-expression. The only way we could accidentally consume one is if a following statement had a leading semicolon, which was semantically important. I don't think is possible in the grammar.

cgrushko:
Sounds good. I can't import the PR before the CLA is signed.

nreid260:
I believe I signed this morning. Perhaps it's still propagating, or Google requires an organization level signature.

PR #276, opened 2022-02-09
Preserve semicolons between call expressions and dead lambdas.

nreid260:
It's possible that removing such semicolons can change the behaviour of the code by reinterpreting the dead lambda as a trailing lambda.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/format/RedundantSemicolonDetector.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/RedundantSemicolonDetector.kt
@@ -17,25 +17,29 @@
 package com.facebook.ktfmt.format
 
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtContainerNodeForControlStructureBody
 import org.jetbrains.kotlin.psi.KtDeclaration
 import org.jetbrains.kotlin.psi.KtEnumEntry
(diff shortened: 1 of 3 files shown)


cgrushko:
This addresses https://github.com/facebookincubator/ktfmt/issues/262 if I'm not mistaken?

cgrushko:
I'm going to import this so it's easier to review and test it on our repo.

core/src/main/java/com/facebook/ktfmt/format/RedundantSemicolonDetector.kt
```diff
@@ -54,12 +58,16 @@ internal class RedundantCommaDetector {
...
-    val prevSibling = element.prevSibling
-    if ((prevSibling is KtIfExpression || prevSibling is KtWhileExpression) &&
+    val prevConcreteSibling = element.getPrevSiblingIgnoringWhitespaceAndComments()
+    if ((prevConcreteSibling is KtIfExpression || prevConcreteSibling is KtWhileExpression) &&
         prevLeaf is KtContainerNodeForControlStructureBody &&
         prevLeaf.text.isEmpty()) {
       return false
     }
+    val nextConcreteSibling = element.getNextSiblingIgnoringWhitespaceAndComments()
+    if (prevConcreteSibling is KtCallExpression && nextConcreteSibling is KtLambdaExpression) {
```
cgrushko, line 68: Can you add a comment with an example that this clause catches? it makes it easier to navigate the code afterwards.

nreid260: Done

PR #279, opened 2022-02-09
Always break get/set definitions onto lines below the property declaration.

nreid260:
This change fixes a crash that occurs on code like `var foo: Int; get() = 0; set(x) { }`

### DIFF
--- a/core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
+++ b/core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
@@ -383,14 +383,32 @@ class FormatterTest {
       |""".trimMargin())
 
   @Test
-  fun `properties with accessors and semicolons on same line`() =
-      assertFormatted(
+  fun `properties with accessors and semicolons on same line`() {
+    val code =
           """
       |class Foo {
       |  var x = false; private set
       |  internal val a by lazy { 5 }; internal get
+      |  var foo: Int; get() = 6; set(x) {};
       |}
-      |""".trimMargin())
+      |""".trimMargin()
+
+    val expected =
+          """
+      |class Foo {
+      |  var x = false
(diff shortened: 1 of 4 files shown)


core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -3002,13 +3024,62 @@ class FormatterTest {
...
+      |  ;
+      |
       |  if (true)  else ;
       |}
       |""".trimMargin()
     assertThatFormatting(code).isEqualTo(expected)
   }
 
+  @Test
+  fun `preserve semicolons between calls and dead lambdas`() {
```
cgrushko, line 3043: leftovers from the previous PR?
(I think you need a different branch for each PR to avoid this)

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -383,14 +383,32 @@ class FormatterTest {
       |""".trimMargin())
 
   @Test
-  fun `properties with accessors and semicolons on same line`() =
-      assertFormatted(
```
cgrushko, line 387: why not keep assertFormatted and change the code? 
we don't need to split to `code` and `expected` in this case, because ktfmt is (supposed to be) determinstic.

cgrushko (changes requested):

cgrushko:
@strulovich , can you take look too?

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
nreid260: Notice that the semicolons are being removed as well. It's important to specify that initial state.

PR #280, opened 2022-02-10
Fix formatting of block-like lambdas (+ scope functions) on long lines.

nreid260:
If everything up to and including the `->` can fit on the current line, then indent only the lambda body. If we must break before the `->`, treat the lambda and scoping function as an expression, and indent the entire thing into a rectangle.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -24,6 +24,7 @@ import com.google.googlejavaformat.Indent
 import com.google.googlejavaformat.Indent.Const.ZERO
 import com.google.googlejavaformat.OpsBuilder
 import com.google.googlejavaformat.Output
+import com.google.googlejavaformat.Output.BreakTag
 import java.util.ArrayDeque
 import java.util.Optional
 import org.jetbrains.kotlin.com.intellij.psi.PsiComment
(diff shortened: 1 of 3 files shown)


cgrushko:
Importing for easier review

cgrushko (changes requested):
Please separate this from https://github.com/facebookincubator/ktfmt/pull/282 and run ktfmt on the code :)

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -875,28 +879,64 @@ class KotlinInputAstVisitor(
...
+     *
+     * We can't denote the lambda (+ scope function) as a block, since (for multiline lambdas) the
+     * rectangle rule would force the entire lambda onto a lower line. Instead, we conditionally
+     * indent all the interior levels of the lambda based on whether we had to break before the
+     * opening brace (or scope function). This mimics the look of a block when the break is taken.
+     *
+     * These conditional indents should not be used inside interior blocks, since that would apply
+     * the condition twice.
+     */
+    val bracePlusBlockIndent = ifBrokeBeforeBrace(blockPlusExpressionBreakIndent, blockIndent)
```
cgrushko, line 913: this shows up both here and in https://github.com/facebookincubator/ktfmt/pull/282....

cgrushko (changes requested):

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
nreid260: #282 is stacked on top of this PR

cgrushko:
Applied the changes to our repo, things look awesome, this is a great fix :) :) :)

PR #281, opened 2022-02-10
Fix crash when type param has an annotation

nreid260:


core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -2501,10 +2501,35 @@ class FormatterTest {
...
-  fun `annotations on functions types parameters`() =
+  fun `annotations in literal function types`() =
       assertFormatted(
           """
-      |val callback: (List<@JvmSuppressWildcards String>) -> Unit = foo
+      |val callback: (@Anno List<@JvmSuppressWildcards String>) -> Unit = foo
+      |""".trimMargin())
+
+  @Test
+  fun `annotations in on type parameters`() =
```
cgrushko, line 2511: nit: is 'in' a typo?

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -1939,21 +1939,7 @@ class KotlinInputAstVisitor(
 
   override fun visitTypeParameter(parameter: KtTypeParameter) {
     builder.sync(parameter)
-    if (parameter.hasModifier(KtTokens.REIFIED_KEYWORD)) {
```
cgrushko, line 1942: ha, that's nice, that we can delete all of this

PR #283, opened 2022-02-13
Fix a bunch of build warnings

nreid260:


cgrushko (changes requested):
There are some build errors reported by CI

cgrushko:
Java 13 still has failures - can you see the logs? (not sure about the permissions here)

nreid260:
I can see them, but I don't understand what would be different. Do you have a minute to take a look? The goal here is just to stop kotlinc from spamming so much.

cgrushko:
No idea; does it build successfully locally for you on Java 13?

nreid260:
Looks like it was some kind of incompatibility with Guava 31

nreid260:
Poke

core/src/main/java/com/facebook/ktfmt/format/KotlinInput.kt
```diff
@@ -116,14 +116,15 @@ class KotlinInput(private val text: String, file: KtFile) : Input() {
...
-    when {
-      length < 0 -> return EMPTY_RANGE
-      length == 0 -> // 0 stands for "format the line under the cursor"
-      length = 1
-    }
+    val expandedLength =
+        when {
+          length < 0 -> return EMPTY_RANGE
+          length == 0 -> 1 // 0 stands for "format the line under the cursor"
+          else -> length
```
cgrushko, line 123: should this be `else -> 1`, at least to preserve the current behavior?

core/src/main/java/com/facebook/ktfmt/format/KotlinInput.kt
```diff
@@ -134,11 +135,11 @@ class KotlinInput(private val text: String, file: KtFile) : Input() {
   }
 
   private fun makePositionToColumnMap(toks: List<KotlinTok>): ImmutableMap<Int, Int> {
-    val builder = ImmutableMap.builder<Int, Int>()
+    val builder = LinkedHashMap<Int, Int>()
     for (tok in toks) {
       builder.put(tok.position, tok.column)
     }
-    return builder.build()
+    return ImmutableMap.copyOf(builder)
```
cgrushko, line 142: what's the build warning here?

core/pom.xml
```diff
@@ -148,7 +148,7 @@
         </dependency>
         <dependency>
             <groupId>org.jetbrains.kotlin</groupId>
-            <artifactId>kotlin-stdlib</artifactId>
+            <artifactId>kotlin-stdlib-jdk7</artifactId>
```
cgrushko, line 151: what's the build warning here?

core/src/main/java/com/facebook/ktfmt/format/KotlinInput.kt
nreid260: This should preserve the current behaviour. `length = 1` isn't a case, it's the behaviour when `length == 0`.

core/src/main/java/com/facebook/ktfmt/format/KotlinInput.kt
nreid260: `ImmutableMap.Builder.build()` throws if there are are duplicate keys. This caused enough confusion that a preferred, explicit method `buildOrThrow` was added in Guava 31.0. Google builds from the newest version of Guava. https://guava.dev/releases/31.0.1-jre/api/docs/com/google/common/collect/ImmutableMap.Builder.html#build()

core/pom.xml
nreid260: I believe this is needed for the changes to core/src/test/java/com/facebook/ktfmt/cli/MainTest.kt

core/src/main/java/com/facebook/ktfmt/format/KotlinInput.kt
cgrushko: /facepalm

core/src/main/java/com/facebook/ktfmt/format/KotlinInput.kt
cgrushko: Got it. I think it's fine this time, but we'll keep running into these problems, since Meta / OSS isn't aware of Guava 31.
Don't know if policy allows that, but you may want to consider adding Guava 29 to third-party (and limiting its `visibility` to ktfmt).

core/src/main/java/com/facebook/ktfmt/format/TypeNameClassifier.kt
```diff
@@ -124,16 +124,16 @@ object TypeNameClassifier {
...
           }
           if (first) {
-            firstUppercase = Character.isUpperCase(element)
+            firstUppercase = Character.isUpperCase(char)
             first = false
           }
-          hasUppercase = hasUppercase or Character.isUpperCase(element)
-          hasLowercase = hasLowercase or Character.isLowerCase(element)
+          hasUppercase = hasUppercase or Character.isUpperCase(char)
+          hasLowercase = hasLowercase or Character.isLowerCase(char)
```
strulovich, line 136: If you're changing this how about using the Kotlin extension methods?

core/src/test/java/com/facebook/ktfmt/cli/MainTest.kt
```diff
@@ -32,7 +33,7 @@ import org.junit.runners.JUnit4
 @RunWith(JUnit4::class)
 class MainTest {
 
-  private val root = createTempDir()
+  private val root = createTempDirectory().toFile()
```
strulovich, line 37: What's the issue here?

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -2069,7 +2069,7 @@ class FormatterTest {
           deduceMaxWidth = true)
 
   @Test
-  fun `handle ? for nullalble types`() =
+  fun `handle qmark for nullalble types`() =
```
strulovich, line 2232: "question mark" please

Let's avoid abbreviations for easier readability

strulovich (changes requested):

core/src/test/java/com/facebook/ktfmt/cli/MainTest.kt
cgrushko: `createTempDir` is marked deprecated

PR #287, opened 2022-02-15
Don't print  op tree by default in tests.

nreid260:
This eliminates a ton of logspam and cuts test run time by 25%.


cgrushko:
Thanks. Didn't realize it has such an affect on run time. 
Question: is there an easy way to tell devs about this option in a failing test? something like, changing https://github.com/facebookincubator/ktfmt/blob/main/core/src/test/java/com/facebook/ktfmt/testutil/KtfmtTruth.kt#L72 to have a custom message in case of expectation failure?

cgrushko:
Actually, I think what we'd want is for the printing to happen automatically if a comparison fails or there's an exception. What do you think?

nreid260:
That's tough to achieve because the op tree is printed by the formatter (multiple times) before the assertions even run. The only solution I could see is to make a debug variant of `format` that also returns the op tree in some form. That form could be an in memory `PrintStream` that holds the string that's eagerly printed today, but I suspect that would cost a similar amount of time.

Additionally, I can say in multiple weeks of programming, the op tree has never been helpful. It's always the AST I want to see.

cgrushko:
Can we rerun formatting if there's an expectation failure, this time with the flag set to true?

nreid260:
All failing tests and only failing tests now print the Op tree.

Printing the tree should not happen in the formatter. Formatting shouldn't have the side-effect of spamming stdout/stderr.

cgrushko (changes requested):
This feels like several PRs merged together.
If we want to avoid printing the op tree unless there's a failure, let's rerun the formatter in https://github.com/facebookincubator/ktfmt/blob/a000e5df7a205edffc7cbd9a19f3e33086a5fc0d/core/src/test/java/com/facebook/ktfmt/testutil/KtfmtTruth.kt#L104 with debuggingPrintOpsAfterFormatting=true (and remove debuggingPrintOpsAfterFormatting=true from other places)
Maybe I'm missing something?

nreid260:
I think it's a bad idea for `format` to print anything to stdout. That's a very strange side effect for a library method.

I think it's a bad idea to re-run the formatter for failing tests. That effectively doubles the execution time of failures.

PR #288, opened 2022-02-15
Noop: refactor KDocWriter to eliminate some dead constructs

nreid260:


cgrushko:
Importing for easier review

PR #289, opened 2022-02-16
Fix crash on trailing comma in arg list after single lambda.

nreid260:


nreid260:
Obsolete by https://github.com/facebookincubator/ktfmt/commit/5e0054a16011e4e24d0892d2583461f94a6fbbb8

PR #290, opened 2022-02-16
Always break before `where` constraints on classes.

nreid260:
This is the simplest reliable fix for the case:
```
class Foo<T> : Bar by bar // There needs to be a line break here or it's a parse error
    where T : Zap
{}
```


nreid260:
Obsolete by https://github.com/facebookincubator/ktfmt/commit/93c7dd96c1e64363d964c036adf84fd6890f9ba4

Issue #291, opened 2022-02-17
Crash when comment separate imports


bethcutler:
```kotlin
import foo
// This comment is pointless
import bar
```

leads to error:
```
Imports not contiguous (perhaps a comment separates them?)
```

We think it is not ideal for the formatter to crash on legal, compilable code.  It's also not good to remove comments, or to attach them to the next line (what if the next line is an unused import that would be removed during formatting?)

We propose sorting all the comments within the import list to the top of the import list.  It may not be exactly what the user intended, but it's not bad.  What do you think?

cgrushko:
We copied GJF's behavior here; has it changed?
I think we shouldn't touch imports at all if we can't do it safely. 
Maybe format the rest of the code? it means we won't delete unused imports.

nreid260:
In the past, I've found that a good way to approach these decisions is to think about what invariants/goals one wants in a tool. I'd propose the following (incomplete) list, which seems to cover all the concerns presented:

1. ktfmt doesn't crash on compilable input
2. ktfmt doesn't delete input information
2. ktfmt output is canonical (including that imports are sorted)
4. **ktfmt output never mixes comments and imports**

Today, we can say that (4) is satisfied, since currently `ktfmt` doesn't produce output when the input has import comments. We sidestep the potentially problematic cases by throwing an error. Unfortunately, that causes a conflict with (1).

@bethcutler proposes a behaviour that satisfies the entire list, so that would seem like an improvement. The question is only, whether we agree about the invariants and their definitions. Of course, there could be more invariants, that I missed, which would constrain the correct behaviour even more.

One potential question about (2) is "what constitutes input information". It could be argued that rearranging comments deletes something, by changing the association between a comment and an import (or range). A solution there is to define away the problem, by saying that `ktfmt` doesn't recognize any such associations; we're then free to move the comments around, as long as we preserve their content. That might seem like cheating, but `ktfmt` does the same thing in many format choices. KDoc layout, whitespace, semicolons, unused imports, all of these are "deleted" because we defined them not to be meaningful.

cgrushko:
Thoughts -
1. My concern is that we'll mess something up. I can imagine somebody trying to add headings to comments (which I think is a bad idea, but). 
2. According to internal data, 97% of times this crash happens it's because of a commented-out import. 

Suggestion: 
1. If the comment is a commented-out import, use @bethcutler 's code, or whatever is simplest (sort as if not a comment, put at the top, put at the bottom, whatever)
[...]

PR #292, opened 2022-02-17
#291: move comments between imports above import list

bethcutler:
Fix for #291


bethcutler:
I addressed this a bit in the #291 issue.  The new behavior is better than the current behavior (these comments would just cause the formatter to crash, so effectively aren't "allowed" right now).  Another option would be to pin the comment to the import just below the comment, but if the formatter removes that import because it is unnecessary, should the comment disappear?  Now the comment may be "attached" to a completely unrelated import.  I feel like this is the safest change we can make, and our users can easily adapt to it.

cgrushko:
Let's first conclude the discussion in https://github.com/facebookincubator/ktfmt/issues/291

PR #298, opened 2022-03-15
Proof of concept for propagating breaks along call expression blocklikes

nreid260:
There's a bad interaction between this change and the current implementation of fluent chain
formatting, however it should give a feel for the effect. I didn't want to spend too much time
polishing this until we had the general idea worked out. It would probably also be smoother
to rework fluent chains first before working on this change.


strulovich:
Can you make this PR not include formatting of the formatter code? That makes it harder to easily review it. I feel ashamed to admit I would have given it more attention if it wasn't the case.

PR #303, opened 2022-03-29
proof of concept editorconfig 

knobo:


knobo:
I suspect this is useless, as ktfmt is suppose to format in accordance with google format.
Unless there is some interest for this, I'll just close the pr and not work any more on it.

knobo:
As I understand it, google-java-format is intentionally not configurable.

cgrushko:
hi! thanks for the interest :) 
As you said, we're trying to minimize the amount of configuration that ktfmt takes, both because similarly to google-java-format we'd prefer only having 1-2 styles to improve readability of Kotlin code everywhere, and because it's harder to maintain high quality when there's an endless number of config combinations.

But, perhaps if we put a single 'style' config in our editorconfig, we can make the command-line, gradle plugin, intellij plugin, all consume the same file, so you only need to change a single place to use a certain style.
Meaning, editorconfig only lets you configure the style (kotlin / java / default), but all tools consume it. 
Anyway, this isn't something that's working right now, but just a thought.

knobo:
So the best solution would be if we would rewrite the whole thing to use 

import and read
- docs/editorconfig/.editorconfig-kotlinlang with --kotlinlang-style
- docs/editorconfig/.editorconfig-dropbox with --dropbox-style
- docs/editorconfig/.editorconfig-google with --google-style

This way we would only have one place to edit the styles. Right now there is two places. One in those files, and one in the code. 

I can understand and appreciate having a tool made for a specific purpose. Like in this case where ktfmt is created to comply with google coding standard. The problem is that this library is used by a generic program (kotlin-language-server), which is supposed to be used by many development environments. It's not your fault that they chose to use ktfmt. Maybe the right solution would to not use ktfmt, since ktfmt is designed to follow google coding style. 

Though, it looks to me as it would be a relative simple and easy change to start using editorconfig, that would make ktfmt more configurable, easier to extend and easier to maintain.

And such it would be beneficial for the whole kotlin community as well.

cgrushko:
I see that ktfmt is used in kotlin-language-server through com.github.fwcd.ktfmt, which is a modified ktfmt maybe?
(btw, super cool they chose ktfmt :) )

If so, com.github.fwcd.ktfmt can expose whatever it needs, even though ktfmt doesn't promise binary stability.

To say so explicitly, we aren't currently planning to expose configuration through either the command line or editorconfig, but you can expose them in e.g., com.github.fwcd.ktfmt (but remember we don't promise binary stability, and APIs may change without notice).

I am open to controlling ktfmt style selection through editorconfig, e.g., `style = kotlinlang` in .editorconfig.

I'll close this PR, given the points above.

Issue #245, opened 2021-07-12
Parameter wrapping incompatible with ktlint with long function parameter


bethcutler:
Current ktfmt formatting:
```kotlin
private fun testWrapping(
  testParameter:
    (
      longEnoughNameToForceMultilineParams: MyVeryLongObject,
      anotherVeryLongName: ALongObject) -> Unit
) {
  // foobar
}
```

What ktlint expects:
```kotlin
private fun testWrapping(
  testParameter: (
    longEnoughNameToForceMultilineParams: MyVeryLongObject,
    anotherVeryLongName: ALongObject
  ) -> Unit
) {
  // foobar
}
```

I think the ktlint formatting looks better.  The ) on a new line would be standard for "google-style".  The ( *not* being on its own line is a bit more compact and it's more obvious what the parens are "attached" to.  However, the current style is similar to the version with a scoping function, e.g.: (Acceptable for both ktfmt and ktlint)
```kotlin
private fun testWrapping(
  testParameter:
    suspend (
      longEnoughNameToForceMultilineParams: MyVeryLongObject,
      anotherVeryLongName: ALongObject,
    ) -> Unit
) {
  // foobar
}
```

bethcutler:
Similar issue:

Original (and ktlint's version, and my personal preference):
```kotlin
typealias Listener = (
  aVeryLongFoo: String,
  anotherVeryLongFoo: String,
  aThirdVeryLongFoo: String,
  aFourthVerylongFoo: String
) -> Unit
```

ktfmt's version (google-style, FWIW):
```kotlin
typealias Listener =
  (
    aVeryLongFoo: String,
    anotherVeryLongFoo: String,
    aThirdVeryLongFoo: String,
    aFourthVerylongFoo: String) -> Unit
```

davidtorosyan:
Fixed by 1bebf0bc4677832a2b7004acf42a938baecd565a, closing.

PR #307, opened 2022-04-07
Add trailing break before closing bracket to comma-separated lists

bethcutler:
Fix for #245


strulovich:
I think the difference for where to place closing parenthesis is one of the things that should be in the Google style, no?

Our style guide keeps `)` in the same line of the last parameter if there's no trailing comma. Could we make this affect the Google style only?

bethcutler:
Yeah, that's an easy enough change to make, so sure.  I'll close this PR and make a new one.

It's unfortunate that the styles are diverging more, though.  Would be nice to keep them more consistent.

PR #311, opened 2022-04-17
Remove use of some deprecated functions

nreid260:


core/src/test/java/com/facebook/ktfmt/cli/ParsedArgsTest.kt
```diff
@@ -132,12 +133,11 @@ class ParsedArgsTest {
   fun `processArgs use the @file option with non existing file`() {
     val out = ByteArrayOutputStream()
 
-    try {
-      ParsedArgs.processArgs(PrintStream(out), arrayOf("@non-existing-file"))
-      fail("expected an exception of type FileNotFoundException but nothing was thrown")
-    } catch (e: FileNotFoundException) {
-      assertThat(e.message).contains("non-existing-file (No such file or directory)")
-    }
+    val e =
```
strulovich, line 136: Please use words as variable names, as we avoid abbreviations to make lines more readable by themselves. I understand that this is conventions dependent, but those are the Ktfmt conventions. This is not a big of an issue, so I might just rename it when pulling since the PR looks 👍

PR #325, opened 2022-06-02
For --google_style, break between ( and long condition expressions

nreid260:
Fixes https://github.com/facebookincubator/ktfmt/issues/319

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -1687,18 +1687,8 @@ class KotlinInputAstVisitor(
   override fun visitWhenExpression(expression: KtWhenExpression) {
     builder.sync(expression)
     builder.block(ZERO) {
-      builder.token("when")
-      expression.subjectExpression?.let { subjectExp ->
-        builder.space()
-        builder.block(ZERO) {
-          builder.token("(")
-          builder.block(if (isGoogleStyle) expressionBreakIndent else ZERO) { visit(subjectExp) }
-          if (isGoogleStyle) {
-            builder.breakOp(Doc.FillMode.UNIFIED, "", ZERO)
-          }
-        }
-        builder.token(")")
-      }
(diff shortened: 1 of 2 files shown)


cgrushko:
Importing for easier review

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -698,6 +698,31 @@ class KotlinInputAstVisitor(
...
+        }
+      } else {
+        builder.block(ZERO) {
+          visit(condition)
+        }
+      }
+    }
+    builder.token(")")
+  }
+
```
strulovich, line 725: We organize methods by chronological order if possible, so caller before callee. Could you move these below all callers?

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -698,6 +698,31 @@ class KotlinInputAstVisitor(
     return extractCallExpression(this)?.lambdaArguments?.isNotEmpty() ?: false
   }
 
+  private fun emitKeywordWithCondition(keyword: String, condition: KtExpression?) {
```
strulovich, line 701: Please add a KDoc. 

We usually state example of the content this parses, so something alike:

```
/** Emits a key word followed by a condition, e.g. `if (b)` or `while (c <d )` */
```

strulovich (changes requested):

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
nreid260: Done

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
nreid260: Done

strulovich (approved):
Thanks!

PR #328, opened 2022-06-08
Prevent adjacent unary operators from merging

nreid260:


core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -2844,6 +2844,27 @@ class FormatterTest {
       |  +4
       |  ++a
       |  --a
+      |
+      |  + +a
+      |  +-a
+      |  -+a
```
cgrushko, line 2850: should this be `+ -a` and `- +a`?  same below for +--a

nreid260: The intent was to test that the extra space is only added when it when two operators would merge.

We could also add a space here, but I'm not sure it makes sense for other combos.

nreid260: It's for unary operators, so the example would be:

```
var a = 1
println(+-a) // -1
```

I agree no human would ever write any of these examples, but a code generators write at least some of them.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -1075,19 +1075,33 @@ class KotlinInputAstVisitor(
...
     builder.block(ZERO) {
-      visit(expression.baseExpression)
-      builder.token(expression.operationReference.text)
+      val baseExpression = expression.baseExpression
+      val operator = expression.operationReference.text
+
+      visit(baseExpression)
+      if (baseExpression is KtPostfixExpression &&
+          baseExpression.operationReference.text.last() == operator.first()) {
+          builder.space()
```
cgrushko, line 1087: is this exercised in any test?

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
cgrushko: got it!

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -1075,19 +1075,33 @@ class KotlinInputAstVisitor(
...
     builder.sync(expression)
     builder.block(ZERO) {
-      builder.token(expression.operationReference.text)
-      visit(expression.baseExpression)
+      val baseExpression = expression.baseExpression
+      val operator = expression.operationReference.text
+
+      builder.token(operator)
+      if (baseExpression is KtPrefixExpression &&
+          operator.last() == baseExpression.operationReference.text.first()) {
```
cgrushko, line 1101: I think this relies on the fact that there are no operators with a mixture of characters, right?
I'm wondering if we should always put a space instead of guessing?

nreid260: That might be better? I like not guessing.

How do you feel about these cases:
```
a--!!
!--a
```

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
nreid260: Not currently. The only example I can think of would be `a!! !!`, but `a!!!!` has the same semantics. Do you have a preference on that case?

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
cgrushko: I hate Kotlin :)
let's add these test-cases?

cgrushko:
Otherwise looks good, thanks!

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
nreid260: Done

PR #330, opened 2022-06-22
Support lambdas and scoping functions in named arguments

nreid260:


core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -1223,6 +1223,38 @@ class FormatterTest {
       |        duration = duration)
       |""".trimMargin())
 
+  @Test
+  fun `named arguments indent scope functions block like`() =
```
davidtorosyan, line 1227: Can you:
1. Add some tests with more than one parameters? Preceding, following, named, not-named, etc.
2. Add some Google format tests
3. Include a "before / after" example in the PR summary? Since this doesn't affect existing tests, it's harder to tell what it does. Unless it does affect existing tests and you haven't updated them yet :P

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -2395,9 +2400,12 @@ class KotlinInputAstVisitor(
     if (isEnabled) {
       open(plusIndent)
     }
-    block()
-    if (isEnabled) {
-      close()
+    try {
+      block()
+    } finally {
```
davidtorosyan, line 2405: This seems like a good change, but I'm not sure how it's useful. Are there cases in the formatter where we swallow exceptions? I'd imagine that if `block()` throws, the whole formatter goes out the window.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -1327,24 +1331,25 @@ class KotlinInputAstVisitor(
...
+        expr.typeArgumentList?.arguments.isNullOrEmpty() &&
+        expr.lambdaArguments.first().getArgumentExpression() is KtLambdaExpression) {
       return true
     }
     return false
   }
 
   /** See [isLambdaOrScopingFunction] for examples. */
-  private fun visitLambdaOrScopingFunction(expr: PsiElement?) {
+  private fun visitLambdaOrScopingFunction(expr: KtExpression?) {
```
davidtorosyan, line 1349: I know you didn't add this function, and this is actually an improvement, but the loose typing + an assertion error seems a little error prone. This probably isn't the place to address it, but good to revisit.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -1327,24 +1331,25 @@ class KotlinInputAstVisitor(
...
     val breakToExpr = genSym()
     builder.breakOp(Doc.FillMode.INDEPENDENT, " ", expressionBreakIndent, Optional.of(breakToExpr))
+    builder.guessToken("*") // In case of a spread argument
```
davidtorosyan, line 1352: nit: if you check `isSpread` instead of guessing the token, the code comments itself :P

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -1037,31 +1037,35 @@ class KotlinInputAstVisitor(
...
       if (hasArgName) {
         visit(argument.getArgumentName())
         builder.space()
         builder.token("=")
-        if (isLambda) {
-          builder.space()
+
+        if (isLambdaOrScopingFunction(argExpr)) {
+          visitLambdaOrScopingFunction(argExpr)
+          return
```
davidtorosyan, line 1050: I'm not crazy about the early return, and if you can find a way to avoid it, that would be great.

I understand if that's not possible, but have you tried messing with the `isEnabled` parameter of `builder.block()`? Basically, I think all you need to do to accomplish what you're going for is disable the block IFF we're looking at a named lambda or scoping function.

Also, you might wanna rebase, because some of this code is changed in mainline.

davidtorosyan (changes requested):
The output looks good, but I have some comments on the implementation, nothing crazy though.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
nreid260: `isSpread` is on `KtArgument` not `KtExpression`. I can't think of a better way to merge these two paths.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
nreid260: It's not about exceptions, but control flow correctness. Because this is method is `inline`, calling `block()` can immediately return, skipping any subsequent statements. The fact that this is possible is a terrible language mistake.

davidtorosyan: Yikes, I had no idea. Good catch.

I just looked it up though and I think adding crossinline to the lambda would help, because it would prevent non-local returns.

I'd be okay with either solution, this one is certainly more flexible. I'd recommend adding a comment about it though.

nreid260: The function could also not be `inline`. If nobody wants that control flow behaviour, there's no value in `inline` here.

davidtorosyan: > there's no value in `inline` here.

There's potentially a perf benefit. I just tried removing the inline, and it increased format time by 1.4%. Not a huge amount (we're talking 3.40ms -> 3.45ms per file, or an increase of 50 microseconds per file), but hey, it's not nothing :)

> I do want that behaviour in the PR though...

I mentioned in another comment, but I think it's possible to not need the early return. Am I understanding right that the early return is to avoid entering the `block` and therefore the linebreaks/indentation that come with it? If so you can just use the `isEnabled` parameter. Let me know if I'm missing something.

nreid260:
I want to continue this chain but I might pause until I can import 0.40 and rebase against that.

hick209:
@nreid260, any status update on this PR?

nreid260:
I do want to eventually merge something like this, but I suspect this one is stale. Basically I want ` = { ... }` in a named args to get the same treatment it would in an assignment, declaration, or expression function.

IIRC the main difficulty was arg around the spread operator `*` which doesn't apply to the currently supported cases and needs to be propagated for args.

I forget how lambdas are formatted in default params, but those should also be made consistent.

hick209:
@nreid260, are you going to work on this in the near future? Else it might be work to close it and open a new one (or reopen this PR) when we're ready to tackle it

PR #336, opened 2022-07-06
Add --stdin-name to specify a file name to use when printing message about stdin input

nreid260:
Closes https://github.com/facebookincubator/ktfmt/issues/335

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/cli/Main.kt
+++ b/core/src/main/java/com/facebook/ktfmt/cli/Main.kt
@@ -70,7 +70,7 @@ class Main(
   fun run(): Int {
     if (parsedArgs.fileNames.isEmpty()) {
       err.println(
-          "Usage: ktfmt [--dropbox-style | --google-style | --kotlinlang-style] [--dry-run] [--set-exit-if-changed] File1.kt File2.kt ...")
+          "Usage: ktfmt [--dropbox-style | --google-style | --kotlinlang-style] [--dry-run] [--set-exit-if-changed] [--stdin-name=<name>] File1.kt File2.kt ...")
       err.println("Or: ktfmt @file")
       return 1
     }
@@ -82,6 +82,9 @@ class Main(
       } catch (e: Exception) {
         1
       }
+    } else if (parsedArgs.stdinName != null) {
+      err.println("Error: --stdin-name can only be used with stdin")
+      return 1
     }
 
     val files: List<File>
@@ -120,7 +123,7 @@ class Main(
    * @return true iff input is valid and already formatted.
    */
   private fun format(file: File?): Boolean {
-    val fileName = file?.toString() ?: "<stdin>"
+    val fileName = file?.toString() ?: parsedArgs.stdinName ?: "<stdin>"
     try {
       val code = file?.readText() ?: BufferedReader(InputStreamReader(input)).readText()
       val formattedCode = Formatter.format(parsedArgs.formattingOptions, code)
@@ -130,7 +133,7 @@ class Main(
       if (file == null) {
         if (parsedArgs.dryRun) {
           if (!alreadyFormatted) {
-            out.println("<stdin>")
+            out.println(fileName)
           }
         } else {
           out.print(formattedCode)
--- a/core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
+++ b/core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
@@ -33,6 +33,8 @@ data class ParsedArgs(
 
     /** Return exit code 1 if any formatting changes are detected. */
     val setExitIfChanged: Boolean,
+    /** File name to report when formating code from stdin */
+    val stdinName: String?,
 ) {
   companion object {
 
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
```diff
@@ -33,6 +33,8 @@ data class ParsedArgs(
 
     /** Return exit code 1 if any formatting changes are detected. */
     val setExitIfChanged: Boolean,
+    /** Name to report when formating code from stdin */
```
cgrushko, line 36: "File name to report when ..."

core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
```diff
@@ -46,10 +48,11 @@ data class ParsedArgs(
 
     /** parseOptions parses command-line arguments passed to ktfmt. */
     fun parseOptions(err: PrintStream, args: Array<String>): ParsedArgs {
-      val fileNames = mutableListOf<String>()
+      var fileNames = mutableListOf<String>()
```
cgrushko, line 51: why the change?

core/src/test/java/com/facebook/ktfmt/cli/MainTest.kt
```diff
@@ -429,4 +439,24 @@ class MainTest {
...
+        Main(
+                emptyInput,
+                PrintStream(out),
+                PrintStream(err),
+                arrayOf("--stdin-name=bar.kt", file.toString()))
+            .run()
+
+    assertThat(file.readText()).isEqualTo(code)
+    assertThat(out.toString("UTF-8")).isEmpty()
+    assertThat(err.toString("UTF-8")).contains("--stdin-name")
```
cgrushko, line 459: use the entire error string? for documentation purposes

core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
```diff
@@ -58,12 +61,14 @@ data class ParsedArgs(
           arg == "--kotlinlang-style" -> formattingOptions = Formatter.KOTLINLANG_FORMAT
           arg == "--dry-run" || arg == "-n" -> dryRun = true
           arg == "--set-exit-if-changed" -> setExitIfChanged = true
+          arg.startsWith("--stdin-name=") -> stdinName = arg.split('=', limit = 2)[1]
```
cgrushko, line 64: add validation here (and tests) - 
if there's no `=` after `--stdin-name`, tell the user how to use it (in particular, this will reduce confusion around `--stdin-name foo.kt`)

does this crash if the user inputs `--stdin-name=` (with nothing after the `=`)?

cgrushko (changes requested):

core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
nreid260: Done.

core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
nreid260: Reverted.

core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
nreid260: Done.

core/src/test/java/com/facebook/ktfmt/cli/MainTest.kt
nreid260: Done.

core/src/main/java/com/facebook/ktfmt/cli/Main.kt
```diff
@@ -120,7 +123,7 @@ class Main(
    * @return true iff input is valid and already formatted.
    */
   private fun format(file: File?): Boolean {
-    val fileName = file?.toString() ?: "<stdin>"
+    val fileName = file?.toString() ?: parsedArgs.stdinName ?: "<stdin>"
```
cgrushko, line 126: if you're in the mood, we can make stdinName = "<stdin>" by default, and make it non-nullable, and then always use it as is

cgrushko:
(I'm waiting for internal review now)

core/src/main/java/com/facebook/ktfmt/cli/Main.kt
nreid260: It might be better if the default value were `""`, since then users could explicitly pass the default value on the CLI.

However, it would be bad if error messaged printed `""` as the file name when it isn't specified. Maybe
```
val fileName = file?.toString() ?: (if (parsedArgs.stdinName.isEmpty()) { "<stdin>" } else { parsedArgs.stdinName })
```

cgrushko: I'm not sure I understand the use-case, but let's not spend more time on this

nreid260: I don't expect to have to pass `""` myself, but I know from other CLIs that its helpful to be able to pass defaults explicitly, particularly when writing shell scripts.

PR #341, opened 2022-07-21
Fix crashes on intersection types (e.g. Foo<A & B>)

nreid260:
Closes https://github.com/facebookincubator/ktfmt/issues/340


core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -231,6 +233,18 @@ class KotlinInputAstVisitor(
...
 
+  /** Example: `A & B`, */
+  override fun visitIntersectionType(type: KtIntersectionType) {
+    builder.sync(type)
+
+    // TODO(strulovich): Should this have the same indentation behaviour as `x && y`?
+    visit(type.getLeftTypeRef())
+    builder.space()
+    builder.token("&")
+    builder.space()
```
cgrushko, line 244: this prevents a line break after the `&` - it's better than crashing, but just wanted to raise this.
Do you want to land this as-is for now, or make it a breakable space first?

nreid260: I'm ok just get something merged, and then worrying about line breaks later . Most uses of intersections are probably with single-letter type variables.

PR #343, opened 2022-07-24
Force-break an indent qualified expression rooted on multiline strings

nreid260:
GJF can't track newlines inside individual tokens, which is how multiline
strings are modeled by the lexer. During printing, this creates a disconnect
between the actual line length, and GJF's internal line length tracking.
This disconnect often triggers the next available break-op, even when the
printed code is below the line length limit.

By forcing a break after every multiline string, the inconsistency is removed.
It is particulalry important for --google-style, where the next natural break
is often between the parens of a chained method call, which forces the closing
paren onto its own line.


nreid260:
Ping.

cgrushko:
Sorry, checking.

PR #345, opened 2022-07-26
Only pretty-print each file one time

nreid260:
The previous formatter flow pretty-printed each file twice, in
order not to change the meaning of the code when removing
semicolons. This made debugging more difficult, because the
second pretty-print was running on already transformed code,
and also caused strange crashes if the result of the first
pretty-print was malformed.

This change eliminates the first pretty-print, running it only
once as the final formatting step. This is done by swapping in
newline charaacters, where appropriate, during semicolon removal.

A side-effect of this change is that a spacing bug is revealed
in the rare case of `if (cond) /** doc */ ;`. Apparently double
pretty-printing hid this behaviour. In exchange, a related bug
is fixed: `if (cond) ; else x` no longer crashes.


nreid260:
Ping

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -3702,6 +3703,7 @@ class FormatterTest {
       |  someLongVariableName.let {
       |    someReallyLongFunctionNameThatMakesThisNotFitInOneLineWithTheAboveVariable()
       |  }
+      |  if (cond)  else 6
```
davidtorosyan, line 3706: Any reason this is two spaces? Is it the same spacing bug as you mentioned with the docstrings?

If so, it's certainly better than crashing, but I'm really curious as to why it's happening.

core/src/main/java/com/facebook/ktfmt/format/RedundantElementRemover.kt
```diff
@@ -65,9 +66,15 @@ object RedundantElementRemover {
...
+      result.replace(element.startOffset, element.endOffset, replacement)
     }
 
     return result.toString()
   }
+
+  private fun PsiElement?.isNewline(): Boolean {
+    if (this !is PsiWhiteSpace) return false
+    return this.text.contains('\n')
+  }
```
davidtorosyan, line 79: Any reason you're using `contains` rather than `==`? Should this instead be `PsiElement?.containsNewline()`?

core/src/main/java/com/facebook/ktfmt/format/RedundantElementRemover.kt
```diff
@@ -65,9 +66,15 @@ object RedundantElementRemover {
             redundantImportDetector.getRedundantImportElements()
 
     for (element in elementsToRemove.sortedByDescending(PsiElement::endOffset)) {
-      result.replace(element.startOffset, element.endOffset, "")
+      val replacement = if (element.nextSibling.isNewline()) "" else "\n"
+      result.replace(element.startOffset, element.endOffset, replacement)
```
davidtorosyan, line 70: Not sure I understand this. Previously, it was removing stuff by replacing tokens with `""`.

Now you're replacing tokens with `"\n"`, unless they're newlines, in which case you remove them? This could use comments to explain what's going on.

davidtorosyan (commented):
Really awesome change, my only complaint is that I don't understand what's happening :P 

Could you explain more about why the new `dropRedundantElements` logic works? And why double formatting was needed in the first place?

core/src/main/java/com/facebook/ktfmt/format/RedundantElementRemover.kt
nreid260: Done.

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
nreid260: Presumably, one space is always inserted after the condition and another before the else.

core/src/main/java/com/facebook/ktfmt/format/RedundantElementRemover.kt
nreid260: Done.

core/src/main/java/com/facebook/ktfmt/format/RedundantElementRemover.kt
davidtorosyan: So the comment is now:
> // Don't insert extra newlines when the semicolon is already a line terminator

Let me see if I got this right:
1. Before, we'd simply remove redundant semicolons
2. Now, we replace them with newlines, *unless* they appear at the end of a line

I'm still not totally clear on how this handles what was previously handled by double formatting, but it's clearly better, so no complaints from me.

nreid260: Before, we would:
  - pretty print, so that no semicolon was in the middle of a line
  - delete the unnecessary semicolons
  - pretty print again

Now we,
  - delete the unnecessary semicolons, which can always be identified
    - inserting a newline if that semicolon was in the middle of a line, to preserve the meaning of the code
  - pretty print once

nreid260:
Awesome! And if we're lucky this is also a noticeable speedup.

davidtorosyan:
> Awesome! And if we're lucky this is also a noticeable speedup.

Good point! I checked and am seeing a 15% speedup, or about 0.5 ms per file. That's pretty awesome!

PR #352, opened 2022-09-21
Fix some regressions in preparation for 0.41

nreid260:
This is a fix for a regression introduced in 8213104a8f1cc0a55800f757486d0e648c8b4ec5


nreid260:
This is the only concerning regression I've seen on our code at 0da0c91349121b271dafd9a13c66dc6c90eb8d03. It's not severe, but is worth fixing.

There's something else about array-literals, but I'm not worried about it.

nreid260:
- Do not indent methods following single-line strings (introduced in 8213104a8f1cc0a55800f757486d0e648c8b4ec5)
- Fix brace spacing of array literal braces

PR #358, opened 2022-10-07
Integrate kdocformatter into ktfmt

cgrushko:


nreid260:
See https://github.com/facebook/ktfmt/pull/359

PR #359, opened 2022-10-09
Merge tnorbye/kdoc-formatter into ktfmt

nreid260:
I've confirmed that this fixes most (if not all) of our known issues with KDoc formatting.

Compared to https://github.com/facebook/ktfmt/pull/358 this version gets the KDoc tests running, and adds copyright headers.

@cgrushko @tnorbye

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/kdoc/CommentType.kt
+++ b/core/src/main/java/com/facebook/ktfmt/kdoc/CommentType.kt
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) Tor Norbye.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.facebook.ktfmt.kdoc
+
+enum class CommentType(
+    /** The opening string of the comment. */
+    val prefix: String,
+    /** The closing string of the comment. */
+    val suffix: String,
+    /**
+     * For multi line comments, the prefix at each comment line after
+     * the first one.
+     */
+    val linePrefix: String
+) {
+    KDOC("/**", "*/", " * "),
+    BLOCK("/*", "*/", ""),
+    LINE("//", "", "// ");
+
+    /**
+     * The number of characters needed to fit a comment on a line: the
+     * prefix, suffix and a single space padding inside these.
+     */
+    fun singleLineOverhead(): Int {
+        return prefix.length + suffix.length + 1 + if (suffix.isEmpty()) 0 else 1
+    }
+
+    /**
+     * The number of characters required in addition to the line comment
+     * for each line in a multi line comment.
+     */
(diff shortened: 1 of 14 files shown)


davidtorosyan:
Thanks for the import @nreid260!

One question for you and @tnorbye, what's the algorithm for `optimal` reflow?

I ask because I lean towards just using greedy unless there's a good reason to use something more complicated - I'm worried about bikeshedding tweaks to the `optimal` algorithm over time.

nreid260:
In testing, optimal seemed to equalize line lengths, which caused short words at line ends to be bumped to the next line. It looked ok/good, but as a user, I'd find it irritating that lines didn't fill.

davidtorosyan:
Agreed. How about we turn off optimal for this integration, and have some follow-up work to remove the relevant code?

Alternatively we can promote it to a CLI option, but unless someone wants to argue for that I don't think it's worth it.

nreid260:
I'm happy to continue using greedy until we have some motivation to change.

tnorbye:
The idea behind optimal  instead of greedy is to avoid raggedness, which makes the text more pleasant looking -- see https://en.wikipedia.org/wiki/Line_wrap_and_word_wrap#Minimum_raggedness .

The overall idea here is to make comments easier to read. I'm not particularly wedded to this; feel free to just go with greedy.

But I do want to make a strong case for the concept of having a maximum comment width *separate* from the line width. That one I really think kttfmt should adopt, even if it's different from the current defaults in ktfmt and java-formatter.

The idea here is that text is *not* like code. For code, you have a lot of nesting, so a long line width makes sense -- 100 characters, 140 characters. It allows deep nesting without the code looking awkward. But most code lines are not 140 characters wide; if they approach the 140 character column, they probably started deeply nested at column 100 with lots of whitespace on the left.

Comments however are not code; they are intended to be read. And very long lines are not readable. That's why books are in portrait orientation, not landscape, and why web sites typically use columns and responsive designs.

If I have a class comment in a file with max-line-width of 140, we should *NOT* flow the text to the full width at 140! Instead, we should have a separate comment width which limits the line widths to something more reasonable. I think a reasonable default is 72 characters.

This means that if you start a comment in code that is nested starting at column 110, then yes, the comment would wrap at column 140 just like the line width says. But if the comment starts further to the left, for example in column 20, then it would wrap at column 92 (20+72), such that each comment line is only 72 characters.

This is why kdocformatter has separate settings for max-line-width and max-comment-width. I really really hope you'll agree to use this setting for ktfmt.

davidtorosyan:
Thanks for the explanation @tnorbye! I'll read about raggedness (from my testing so far I'm guessing we'll still go with greedy though) and I'll do some testing with the line length - I hadn't even thought of it.

cgrushko:
@tnorbye, in the context of tracking upstream, how do you see kdocformatter going forward?
For example, if it continues existing as an independent library, we might want to make it easier to track it (e.g., keeping the original package name, placing in a subdir named after the version, etc.).

cgrushko:
Re greedy vs optimal: I suggest we use whatever is similar to what we have now, and change to optimal in a subsequent PR.

Re comment width: interesting! I need to think about this :) I also think there's a difference between kdoc and inline comments, but I'd like some consistency. For now, again I suggest we stick to what looks similar to what we have now. (meaning, comment width == code width)

davidtorosyan:
Going ahead with landing, here's the follow-ups:
1. We'll evaluate optimal vs greedy, and either (a) switch to optimal, (b) remove the optimal code.
2. I'll likely set `allowParamBrackets` to `false` for consistency
3. We'll expose `maxCommentWidth` as a parameter to the ktfmt CLI

@tnorbye / @nreid260 let us know if there's anything else we're missing. I'd like it if `ktfmt` can be used wherever `kdocformatter` was used before.

PR #372, opened 2022-12-09
Miscellaneous fixes to the KDoc formatter

tnorbye:
This CL fixes a number of bugs, covered by associated unit tests:
   - Line overrun when using closed-open interval notation
   - More gracefully handle unterminated [] references (for example when comment is using it in things like [closed, open) intervals)
   - Recognize and convert accidentally capitalized kdoc tags like @See
   - If you have a [ref] which spans a line such that the # ends up as a new line, don't treat this as a "# heading".
   - Make sure we don't line break at an expression starting with ">" since that would turn into a quoted line.
   - If you're using optimal line breaking and there's a really long, unbreakable word in the paragraph, switch that paragraph over to greedy line breaking (to make the paragraph better balanced since the really long word throws the algorithm off.)
   - Fix a few scenarios where markup conversion from <p> and </p> wasn't converting everything.
   - Allow @property[name], not just @param[name]
   - The --hanging-indent flag now also sets the nested list indent (if >= 3)
- Some minor code cleanup.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/kdoc/KDocFormattingOptions.kt
+++ b/core/src/main/java/com/facebook/ktfmt/kdoc/KDocFormattingOptions.kt
@@ -60,7 +60,8 @@ class KDocFormattingOptions(
 
   /**
    * How many spaces to use for hanging indents in numbered lists and after block tags. Using 4 or
-   * more here will result in subsequent lines being interpreted as block formatted.
+   * more here will result in subsequent lines being interpreted as block formatted by IntelliJ (but
+   * not Dokka).
    */
   var hangingIndent: Int = 2
 
--- a/core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
+++ b/core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
@@ -101,10 +101,6 @@ class Paragraph(private val task: FormattingTask) {
     return content.isEmpty()
   }
 
-  private fun hasClosingPre(): Boolean {
-    return content.contains("</pre>", ignoreCase = false) || next?.hasClosingPre() ?: false
-  }
-
   fun cleanup() {
     val original = text
 
@@ -164,11 +160,23 @@ class Paragraph(private val task: FormattingTask) {
   }
 
   private fun convertMarkup(s: String): String {
-    if (s.none { it == '<' || it == '&' || it == '{' }) return s
+    // Whether the tag starts with a capital letter and needs to be cleaned, e.g. `@See` -> `@see`.
+    // (isKDocTag only allows the first letter to be capitalized.)
+    val convertKDocTag = s.isKDocTag() && s[1].isUpperCase()
+
+    if (!convertKDocTag && s.none { it == '<' || it == '&' || it == '{' }) {
+      return s
+    }
 
     val sb = StringBuilder(s.length)
     var i = 0
     val n = s.length
+
+    if (convertKDocTag) {
+      sb.append('@').append(s[1].lowercaseChar())
+      i += 2
+    }
+
     var code = false
     var brackets = 0
     while (i < n) {
@@ -335,8 +343,12 @@ class Paragraph(private val task: FormattingTask) {
     return reflow(words, lineWidth, hangingIndentSize)
   }
 
-  fun reflow(words: List<String>, lineWidth: Int, hangingIndentSize: Int): List<String> {
-    if (options.alternate || !options.optimal || hanging && hangingIndentSize > 0) {
+  private fun reflow(words: List<String>, lineWidth: Int, hangingIndentSize: Int): List<String> {
+    if (options.alternate ||
+        !options.optimal ||
+        hanging && hangingIndentSize > 0 ||
+        // An unbreakable long word may make other lines shorter and won't look good
+        words.any { it.length > lineWidth }) {
       // Switch to greedy if explicitly turned on, and for hanging indent
       // paragraphs, since the current implementation doesn't have support
       // for a different maximum length on the first line from the rest
@@ -405,7 +417,8 @@ class Paragraph(private val task: FormattingTask) {
         word.startsWith("```") ||
         word.isDirectiveMarker() ||
         word.startsWith("@") || // interpreted as a tag
-        word.isTodo()) {
+        word.isTodo() ||
+        word.startsWith(">")) {
       return false
     }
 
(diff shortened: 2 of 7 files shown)


core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
```diff
@@ -101,10 +101,6 @@ class Paragraph(private val task: FormattingTask) {
     return content.isEmpty()
   }
 
-  private fun hasClosingPre(): Boolean {
-    return content.contains("</pre>", ignoreCase = false) || next?.hasClosingPre() ?: false
-  }
```
strulovich, line 106: Style wise we go with the IntelliJ recommendation to use `booleanNullable == true` instead of `booleanNullable ?: false`

core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
```diff
@@ -164,11 +160,19 @@ class Paragraph(private val task: FormattingTask) {
   }
 
   private fun convertMarkup(s: String): String {
-    if (s.none { it == '<' || it == '&' || it == '{' }) return s
+    val convertKDocTag = s.isKDocTag() && s[1].isUpperCase()
```
strulovich, line 163: Should this deal with any kind of uppercase letters in the tag? Or are other capitalized tags not expected as well?

In any case, I would give an example based comment above (and I'm not a big comment fan) such as:
```
// Whether the tag starts with a capital letter and needs to be cleaned, e.g. `@See` -> `@see`
```

core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
```diff
@@ -335,8 +339,13 @@ class Paragraph(private val task: FormattingTask) {
     return reflow(words, lineWidth, hangingIndentSize)
   }
 
-  fun reflow(words: List<String>, lineWidth: Int, hangingIndentSize: Int): List<String> {
-    if (options.alternate || !options.optimal || hanging && hangingIndentSize > 0) {
+  private fun reflow(words: List<String>, lineWidth: Int, hangingIndentSize: Int): List<String> {
+    if (options.alternate ||
+        !options.optimal ||
+        hanging && hangingIndentSize > 0 ||
+        // If we have an unbreakably long word this tends to make the optimal
```
strulovich, line 346: nit: `unbreakable`

But also, personally I think this comment can be written with less words to the same effect: (unless I'm missing some extra nuance)
```
// An unbreakable long word may make other lines shorter and won't look good
```

core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
```diff
@@ -164,11 +160,19 @@ class Paragraph(private val task: FormattingTask) {
   }
 
   private fun convertMarkup(s: String): String {
-    if (s.none { it == '<' || it == '&' || it == '{' }) return s
+    val convertKDocTag = s.isKDocTag() && s[1].isUpperCase()
+
+    if (!convertKDocTag && s.none { it == '<' || it == '&' || it == '{' }) return s
```
strulovich, line 165: style consistency nit: we use braces on ifs even when single line

core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
```diff
@@ -438,39 +455,53 @@ class Paragraph(private val task: FormattingTask) {
...
+            from + 2
+          } else {
+            from + 1
+          }
+      var to = words.size
+      for (i in start until words.size) {
+        val next = words[i]
+        if (next.startsWith("[") && !next.startsWith("[[")) {
+          // find end
+          var f = -1
```
strulovich, line 472: Can we give this a better name than `f`?

core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
```diff
@@ -438,39 +455,53 @@ class Paragraph(private val task: FormattingTask) {
...
+          for (k in i until words.size) {
```
strulovich, line 473: Usually the next index variable used after `i` is `j`, no?

core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
```diff
@@ -438,39 +455,53 @@ class Paragraph(private val task: FormattingTask) {
...
+            if (words[k].contains("]")) {
```
strulovich, line 474: Use `in` syntax?

core/src/main/java/com/facebook/ktfmt/kdoc/KDocFormattingOptions.kt
```diff
@@ -60,7 +60,8 @@ class KDocFormattingOptions(
 
   /**
    * How many spaces to use for hanging indents in numbered lists and after block tags. Using 4 or
-   * more here will result in subsequent lines being interpreted as block formatted.
+   * more here will result in subsequent lines being interpreted as block formatted by IntelliJ (not
+   * Dokka).
```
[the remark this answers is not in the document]
tnorbye: I've made this change. 

I have very little experience with github and trying to amend my own CL (instead of creating new CLs on top of it; not sure what's recommended here; I normally use gerrit where amending to incorporate comments is the recommended way.) I tried to do a force push (--force-with-lease) which I read somewhere is the correct way to do this, but not sure if I did it correctly!

core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
tnorbye: Thanks, will try to remember that! (This code is deleted by my CL so nothing to do here I assume.)

core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
tnorbye: I used your suggestion (and added that the first-letter only restriction is made by isKDocTag. I thought I'd be conservative here and only look for (accidentally?) capitalized tags, not making them case insensitive.

core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
tnorbye: Thanks, I used your suggestion.

core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
tnorbye: I've updated it to do that. I'm assuming this is the kind of change ktfmt would do if I knew how to run the right formatting hook on the source code :-)

core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
tnorbye: Yes, should have been `j`. I know that's not a great name either, but these are basically indices used to scan ahead in a nested context.

core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
tnorbye: Yes, that's what I changed it to.

core/src/main/java/com/facebook/ktfmt/kdoc/Paragraph.kt
tnorbye: Since you suggested it I made the switch, but, ... I don't think I like it; I think "String.contains(substring)" is more natural. Maybe it's just habit with years and years of string manipulation...  :-)

PR #378, opened 2023-01-21
De-inline OpsBuilder.block()

nreid260:
As implemented, this function is unsafe to inline. Outer return from the callback would skip the call to builder.close().


nreid260:
Preparation for https://github.com/facebook/ktfmt/issues/380

PR #379, opened 2023-01-21
Eliminate mis-paired OpsBuilder.open()/close()

nreid260:
These explicit calls are misleading because they pair with the calls inside OpsBuilder.block(), so the indention of the source doesn't match the block structure of the output.

I believe the only potentially affected case is unnamed function-literal-type params, which cannot have initializers, making this a no-op.


nreid260:
Preparation for https://github.com/facebook/ktfmt/issues/380

Issue #380, opened 2023-01-22
Comments after assignment prevents ktfmt indentation 


nreid260:
Before (correct)

```
package test

val foo =
  // This is a comment relative to fooBar
  fooBar {}
```

After (wrong)

```
package test

val foo =
// This is a comment relative to fooBar
fooBar {}
```

nreid260:
I have a fix for this as well, in a few parts. Can anyone coordinate on the PRs with me?

PR #381, opened 2023-01-27
Create FenceCommentsOp and use it to simplify some confusing comment handling situations

nreid260:


nreid260:
https://github.com/facebook/ktfmt/issues/380

cgrushko:
@nreid260  thanks for the PR. Is this also relevant to GJF?

nreid260:
This might be useful, I'll see what the upstream owners think. We can always revert this later if it lands in gjf.

core/src/main/java/com/facebook/ktfmt/format/FenceCommentsOp.kt
```diff
@@ -0,0 +1,21 @@
+package com.facebook.ktfmt.format
```
cgrushko, line 1: @nreid260 , this file needs a copyright header.
But I don't know what to write there 🤷 I think the easiest is to put both Google and Meta (i.e., two stacked headers)

nreid260: Added header. I don't see this as particularly different than code I've written in other files.

PR #385, opened 2023-02-17
Enforce blank lines between class members

nreid260:
This change is effectively a separation of the behaviour for block-bodies vs block-expressions. In doing so, several simplifications were uncovered.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -45,6 +45,7 @@ import org.jetbrains.kotlin.psi.KtCallExpression
 import org.jetbrains.kotlin.psi.KtCallableReferenceExpression
 import org.jetbrains.kotlin.psi.KtCatchClause
 import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtClassBody
 import org.jetbrains.kotlin.psi.KtClassInitializer
 import org.jetbrains.kotlin.psi.KtClassLiteralExpression
 import org.jetbrains.kotlin.psi.KtClassOrObject
@@ -170,10 +171,9 @@ class KotlinInputAstVisitor(
           true,
           function.valueParameterList,
           function.typeConstraintList,
-          function.bodyBlockExpression,
-          function.bodyExpression,
+          function.bodyBlockExpression ?: function.bodyExpression,
           function.typeReference,
-          function.bodyBlockExpression?.lBrace != null)
+      )
     }
   }
 
@@ -289,10 +289,8 @@ class KotlinInputAstVisitor(
       emitParenthesis: Boolean,
       parameterList: KtParameterList?,
       typeConstraintList: KtTypeConstraintList?,
-      bodyBlockExpression: KtBlockExpression?,
-      nonBlockBodyExpressions: KtExpression?,
+      bodyExpression: KtExpression?,
       typeOrDelegationCall: KtElement?,
-      emitBraces: Boolean
   ) {
     builder.block(ZERO) {
       if (modifierList != null) {
@@ -361,19 +359,19 @@ class KotlinInputAstVisitor(
         builder.space()
         visit(typeConstraintList)
       }
-      if (bodyBlockExpression != null) {
+      if (bodyExpression is KtBlockExpression) {
         builder.space()
-        visitBlockBody(bodyBlockExpression, emitBraces)
-      } else if (nonBlockBodyExpressions != null) {
+        visit(bodyExpression)
+      } else if (bodyExpression != null) {
         builder.space()
         builder.block(ZERO) {
           builder.token("=")
-          if (isLambdaOrScopingFunction(nonBlockBodyExpressions)) {
-            visitLambdaOrScopingFunction(nonBlockBodyExpressions)
+          if (isLambdaOrScopingFunction(bodyExpression)) {
+            visitLambdaOrScopingFunction(bodyExpression)
           } else {
             builder.block(expressionBreakIndent) {
               builder.breakOp(Doc.FillMode.INDEPENDENT, " ", ZERO)
-              builder.block(ZERO) { visit(nonBlockBodyExpressions) }
+              builder.block(ZERO) { visit(bodyExpression) }
             }
           }
         }
@@ -389,23 +387,22 @@ class KotlinInputAstVisitor(
     return Output.BreakTag()
   }
 
-  private fun visitBlockBody(bodyBlockExpression: PsiElement, emitBraces: Boolean) {
-    if (emitBraces) {
-      builder.token("{", Doc.Token.RealOrImaginary.REAL, blockIndent, Optional.of(blockIndent))
-    }
+  private fun emitBracedBlock(
+      bodyBlockExpression: PsiElement,
+      emitChildren: (Array<PsiElement>) -> Unit,
+  ) {
+    builder.token("{", Doc.Token.RealOrImaginary.REAL, blockIndent, Optional.of(blockIndent))
     val statements = bodyBlockExpression.children
     if (statements.isNotEmpty()) {
       builder.block(blockIndent) {
         builder.forcedBreak()
         builder.blankLineWanted(OpsBuilder.BlankLineWanted.PRESERVE)
-        visitStatements(statements)
+        emitChildren(statements)
       }
       builder.forcedBreak()
       builder.blankLineWanted(OpsBuilder.BlankLineWanted.NO)
     }
-    if (emitBraces) {
-      builder.token("}", blockIndent)
-    }
+    builder.token("}", blockIndent)
   }
 
   private fun visitStatement(statement: PsiElement) {
@@ -1384,10 +1381,9 @@ class KotlinInputAstVisitor(
                 accessor.bodyExpression != null || accessor.bodyBlockExpression != null,
                 accessor.parameterList,
                 null,
-                accessor.bodyBlockExpression,
-                accessor.bodyExpression,
+                accessor.bodyBlockExpression ?: accessor.bodyExpression,
                 accessor.returnTypeReference,
-                accessor.bodyBlockExpression?.lBrace != null)
+            )
           }
         }
       }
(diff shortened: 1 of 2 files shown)


cgrushko:
What's GJF's behavior for fields?

nreid260:
> What's GJF's behavior for fields?

GJF preserves up to 1 blank line between field declarations.

cgrushko:
> I would support that behaviour for simple property declarations, but not when there are getter/setter definitions.

I agree. 

For many fields (properties) one after the other, forcing newlines seems too much. WDYT?

nreid260:
I appreciate the consistency of "one blank line between each declaration". I also like that it means humans no longer have to chose (I remember spending time debating whether fields were "related enough" to be grouped, and then looking at the code a month later, being to totally indifferent to the choice I made).

However, if you feel like implementing it as part of this merge, that's also fine be me. The main thing is that methods *definitely* should be spaced.

cgrushko:
I totally agree about methods and properties with getters/setters. I'm only concerned about the amount of vertical space it'll add to 'simple' properties.

nreid260:
Do you have an example of a class where spacing the properties makes things worse?

davidtorosyan:
> Do you have an example of a class where spacing the properties makes things worse?

I did some ripgrep'ing through Kotlin's own library to find this example: [kotlin/IntrinsicType.kt](https://github.com/JetBrains/kotlin/blob/master/kotlin-native/runtime/src/main/kotlin/kotlin/native/internal/IntrinsicType.kt)

```kt
        // Coroutines
        const val GET_CONTINUATION              = "GET_CONTINUATION"
        const val RETURN_IF_SUSPENDED           = "RETURN_IF_SUSPENDED"

        // Interop
        const val INTEROP_READ_PRIMITIVE        = "INTEROP_READ_PRIMITIVE"
        const val INTEROP_WRITE_PRIMITIVE       = "INTEROP_WRITE_PRIMITIVE"
        const val INTEROP_GET_POINTER_SIZE      = "INTEROP_GET_POINTER_SIZE"
        const val INTEROP_NATIVE_PTR_TO_LONG    = "INTEROP_NATIVE_PTR_TO_LONG"
        const val INTEROP_NATIVE_PTR_PLUS_LONG  = "INTEROP_NATIVE_PTR_PLUS_LONG"
        const val INTEROP_GET_NATIVE_NULL_PTR   = "INTEROP_GET_NATIVE_NULL_PTR"
        const val INTEROP_CONVERT               = "INTEROP_CONVERT"
        const val INTEROP_BITS_TO_FLOAT         = "INTEROP_BITS_TO_FLOAT"
        const val INTEROP_BITS_TO_DOUBLE        = "INTEROP_BITS_TO_DOUBLE"
        const val INTEROP_SIGN_EXTEND           = "INTEROP_SIGN_EXTEND"
        const val INTEROP_NARROW                = "INTEROP_NARROW"
        const val INTEROP_STATIC_C_FUNCTION     = "INTEROP_STATIC_C_FUNCTION"
        const val INTEROP_FUNPTR_INVOKE         = "INTEROP_FUNPTR_INVOKE"

        // Worker
        const val WORKER_EXECUTE                = "WORKER_EXECUTE"

```

I'd argue this would look worse with uniform newlines everywhere, as opposed to these logical groupings.

>  I also like that it means humans no longer have to chose (I remember spending time debating whether fields were "related enough" to be grouped

I don't know if I agree this should be outsourced. I go through a similar thought process when it comes to property naming, method ordering, etc. Basically, just because it's fuzzy doesn't mean there isn't meaningful information in there.

nreid260:
I feel like that missed an operative part of the sentence:

> and then looking at the code a month later, being to totally indifferent to the choice I made

Regarding [kotlin/IntrinsicType.kt](https://github.com/JetBrains/kotlin/blob/master/kotlin-native/runtime/src/main/kotlin/kotlin/native/internal/IntrinsicType.kt), `ktfmt` is already going to shatter that file. That beautiful handcrafted spacing won't survive 😂 

Overall, I don't have a strong opinion on whether property declarations should have separating lines, but in the absence of a strong opinion from someone, I advocate against adding complexity.

davidtorosyan:
> I feel like that missed an operative part of the sentence:

Fair enough, but I think my point stands. Sometimes you'll end up indifferent, sometimes not. That's the fuzzy part.

> Regarding [kotlin/IntrinsicType.kt](https://github.com/JetBrains/kotlin/blob/master/kotlin-native/runtime/src/main/kotlin/kotlin/native/internal/IntrinsicType.kt), ktfmt is already going to shatter that file. That beautiful handcrafted spacing won't survive 😂

I just ran it through [the online formatter](https://facebook.github.io/ktfmt/) and got this:
```kt
    // Coroutines
    const val GET_CONTINUATION = "GET_CONTINUATION"
    const val RETURN_IF_SUSPENDED = "RETURN_IF_SUSPENDED"

    // Interop
    const val INTEROP_READ_PRIMITIVE = "INTEROP_READ_PRIMITIVE"
    const val INTEROP_WRITE_PRIMITIVE = "INTEROP_WRITE_PRIMITIVE"
    const val INTEROP_GET_POINTER_SIZE = "INTEROP_GET_POINTER_SIZE"
    const val INTEROP_NATIVE_PTR_TO_LONG = "INTEROP_NATIVE_PTR_TO_LONG"
    const val INTEROP_NATIVE_PTR_PLUS_LONG = "INTEROP_NATIVE_PTR_PLUS_LONG"
    const val INTEROP_GET_NATIVE_NULL_PTR = "INTEROP_GET_NATIVE_NULL_PTR"
    const val INTEROP_CONVERT = "INTEROP_CONVERT"
    const val INTEROP_BITS_TO_FLOAT = "INTEROP_BITS_TO_FLOAT"
    const val INTEROP_BITS_TO_DOUBLE = "INTEROP_BITS_TO_DOUBLE"
    const val INTEROP_SIGN_EXTEND = "INTEROP_SIGN_EXTEND"
    const val INTEROP_NARROW = "INTEROP_NARROW"
    const val INTEROP_STATIC_C_FUNCTION = "INTEROP_STATIC_C_FUNCTION"
    const val INTEROP_FUNPTR_INVOKE = "INTEROP_FUNPTR_INVOKE"

    // Worker
    const val WORKER_EXECUTE = "WORKER_EXECUTE"
```

I think "shattered" is a bit much, it's still perfectly readable. Maybe we should look into preserving space alignment though...

> Overall, I don't have a strong opinion on whether property declarations should have separating lines, but in the absence of a strong opinion from someone, I advocate against adding complexity.

I can't tell if you're suggesting we continue with your suggestion of adding separating lines or not 😛

nreid260:
So, should I interpret this as a hard requirement that properties can be on adjacent lines? I don't really want to implement that, but I will if it means getting a fix for spacing between adjacent methods.

I note that there's a big-ish set of cases to consider: every pair of [prop, prop with comment, prop with gettter/setter, prop with annotation, other declaration]

cgrushko:
The several folks I asked about this change favored GJF's behavior.
How do others at Google feel about this?

davidtorosyan:
> So, should I interpret this as a hard requirement that properties can be on adjacent lines?

Yeah what Carmi said, less of a hard requirement and more trying to understand what looks best. Especially since this would impact a lot of already formatted code, the bar is a bit higher. 

For the previous example I gave, that was just in response to "what's a case where this looks worse?". It's still subjective but that's where we're coming from.

nreid260:
Preserved blank lines between properties.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -389,37 +388,39 @@ class KotlinInputAstVisitor(
...
 
   private fun visitStatement(statement: PsiElement) {
     builder.block(ZERO) { visit(statement) }
     builder.guessToken(";")
   }
 
-  private fun visitStatements(statements: Array<PsiElement>) {
+  private fun visitStatements(
+      statements: Array<PsiElement>,
+      linesBetweenStatements: OpsBuilder.BlankLineWanted,
```
davidtorosyan, line 416: Why make this a parameter? From what I can tell you're always passing in `PRESERVE` here.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -389,37 +388,39 @@ class KotlinInputAstVisitor(
     return Output.BreakTag()
   }
 
-  private fun visitBlockBody(bodyBlockExpression: PsiElement, emitBraces: Boolean) {
-    if (emitBraces) {
-      builder.token("{", Doc.Token.RealOrImaginary.REAL, blockIndent, Optional.of(blockIndent))
-    }
+  private fun emitBracedBlock(
```
davidtorosyan, line 391: I'm having trouble understanding if the changes related to `emitBracedBlock` are pure refactoring or behavioral changes.

If they're refactoring, the changes look good, but it'd be easier to review as a separate PR. I know we can take a long time to review, but a no-op clean-up should be pretty quick. It's fine for this change, just a note for the future. At the least, calling it out in the PR summary would be nice.

If they're behavioral, could use a description of the changes.

That said, I'm 90% sure this is a good refactor.

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -6495,6 +6508,106 @@ class FormatterTest {
...
+      |}
+      |"""
+            .trimMargin()
+
+    val expected =
+        """
+      |class Foo {
+      |  val x = 0
+      |  val x = 0
+      |    private get
```
davidtorosyan, line 6593: My code suggestion would force a newline between these two. Personally I think that would make sense to separate.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -1897,9 +1848,58 @@ class KotlinInputAstVisitor(
...
+          prev is KtProperty -> {
+            when {
+              prev.getter != null -> builder.blankLineWanted(OpsBuilder.BlankLineWanted.YES)
+              prev.setter != null -> builder.blankLineWanted(OpsBuilder.BlankLineWanted.YES)
+              curr is KtProperty -> builder.blankLineWanted(OpsBuilder.BlankLineWanted.PRESERVE)
+              else -> builder.blankLineWanted(OpsBuilder.BlankLineWanted.YES)
+            }
+          }
+          else -> builder.blankLineWanted(OpsBuilder.BlankLineWanted.YES)
+        }
```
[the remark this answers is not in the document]
davidtorosyan: I see from your unit test that (1) is intentional, my vote is to add a newline for these cases, but don't feel strongly about it.

davidtorosyan (commented):
Left some comments! Overall I'm happy with this, just some clarification questions and style nits.

From the code samples I've seen I think this an improvement (now that simple properties can be next to each other).

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -16,6 +16,7 @@
 
 package com.facebook.ktfmt.format
 
+import org.jetbrains.kotlin.kdoc.psi.api.KDoc
```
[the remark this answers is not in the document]
nreid260: Done.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -1897,9 +1848,58 @@ class KotlinInputAstVisitor(
...
+            if (builder.peekToken() == Optional.of(",")) {
+              builder.token(",")
+              builder.forcedBreak()
+            }
+          }
+        }
+        builder.guessToken(";")
+      }
+
+      if (enumEntries.isNotEmpty() && nonEnumEntryMembers.isNotEmpty()) {
```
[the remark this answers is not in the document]
nreid260: Done.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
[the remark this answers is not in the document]
nreid260: How's this?

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
nreid260: Reverted.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
nreid260: It's deleting unused execution paths. Now that the function only emits braced blocks, the braces are unconditional.

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
nreid260: That would frustrate the case of a private backing property, which is the strongest form of grouping

```
private var _foo: Foo?
val foo: Foo
  get() = _foo!!
```

davidtorosyan: Got it, makes sense.

PR #392, opened 2023-03-08
Don't crash on nullable reference types with parens

nreid260:


cgrushko:
Guessing tokens is problematic because the operation can cross grammar nodes. Unfortunately I can't recall the exact bug we had that was caused by this.

But it goes something like this - if `visitNullableType()` is called from another node, consumes a closing paren `)`, but that closing parenthesis belongs to the outer node, then we'll break formatting.
If I had a concrete example it would've been easier.

It's possible that in this specific case this breakage can't happen, but I don't have a good mental model to say why.

So - let's use guessToken only if we absolutely can't predict whether parenthesis should be present or not.

nreid260:
Done

nreid260:
Ping

PR #396, opened 2023-04-14
Allow optional break after string-literals

nreid260:


core/src/test/java/com/facebook/ktfmt/testutil/KtfmtTruth.kt
```diff
@@ -68,7 +68,7 @@ fun assertFormatted(
 
 fun assertThatFormatting(code: String): FormattedCodeSubject {
   fun codes(): Subject.Factory<FormattedCodeSubject, String> {
-    return Subject.Factory { metadata, subject -> FormattedCodeSubject(metadata, subject) }
+    return Subject.Factory { metadata, subject -> FormattedCodeSubject(metadata, subject!!) }
```
nreid260, line 71: The API of Subject.Factory is changing (has changed within Google) so that the `subject` is nullable even if the constructor doesn't accept null.

PR #398, opened 2023-04-26
Never insert a linebreak between `@sample very.long.name`

nreid260:
Inserting linebreaks here causes problems for Android Studio and likely other tools. This was presumed fixed when kdoc-formatter was merged, but apparently it had the same beahviour.

The fix will be upstreamed to kdoc-formatter as well

Closes https://github.com/facebookincubator/ktfmt/issues/310


nreid260:
@tnorbye

Issue #374, opened 2022-12-11
Format about context receiver


danny-su:
The empty line between context declaration and function definition should be removed.
<img width="390" alt="image" src="https://user-images.githubusercontent.com/12178855/206890569-a450007c-694d-4c28-892f-f8e023627b4a.png">

JavierSegoviaCordoba:
Duplicated by https://github.com/facebook/ktfmt/issues/314

Issue #314, opened 2022-04-22
context receivers support


JavierSegoviaCordoba:
Currently, `context` add an unnecessary breakline:

```kotlin
context(ScanResult)

internal val ClassInfo.contextClasses: List<KClass<out Any>>
  get() =
    getClassesWithAnnotation(this.name).map { contextClassInfo ->
      contextClassInfo.loadClass().kotlin
    }
```

should be:

```kotlin
context(ScanResult)
internal val ClassInfo.contextClasses: List<KClass<out Any>>
  get() =
    getClassesWithAnnotation(this.name).map { contextClassInfo ->
      contextClassInfo.loadClass().kotlin
    }
```

strulovich:
It is likely we will only fix this when this comes out of preview. Happy to accept a PR once the dependencies sort themselves out (though it looks it might be future compatible from the fact Ktfmt doesn't crash on this)

JavierSegoviaCordoba:
I would like to know how it will be combined with annotations.

### 1
```kotlin
context(Persistence)
@Provider class Repo
```
### 2

```kotlin
context(Persistence)
@Provider
class Repo
```
### 3

```kotlin
@Provider
context(Persistence)
class Repo
```

The first one is which I like more but if there are a lot of annotations it will not look very good, so I think 3 is the best one.

JavierSegoviaCordoba:
@strulovich I tried this myself but after updating Kotlin to 1.7.10, I cannot get the context receivers in `visitNamedFunction`, I got an empty list even though the `KotlinInput` shows clearly they are there.

Not sure if I am missing something.

JavierSegoviaCordoba:
@strulovich as they are not going to be on 1.8.0, can this be prioritized a bit, please?

PR #400, opened 2023-04-29
Support context receivers

### LINKED ISSUE #397: Context receivers give `ParseError`
Using `com.diffplug.spotless:spotless-plugin-gradle:6.18.0` configured to use `ktfmt().kotlinlangStyle()` fails to parse usage of context receivers.

Minimum example:

```kotlin
class A {
    context(String)
    fun test() {}
}
```

leads to
```
Step 'ktfmt' found problem in 'A.kt':
2:5: error: Expecting member declaration
com.facebook.ktfmt.format.ParseError: 2:5: error: Expecting member declaration
	at com.facebook.ktfmt.format.Parser.throwParseError(Parser.kt:71)
	at com.facebook.ktfmt.format.Parser.parse(Parser.kt:65)
	at com.facebook.ktfmt.format.Formatter.sortedAndDistinctImports(Formatter.kt:143)
	at com.facebook.ktfmt.format.Formatter.format(Formatter.kt:90)
	at com.diffplug.spotless.glue.ktfmt.KtfmtFormatterFunc.apply(KtfmtFormatterFunc.java:54)
	at com.diffplug.spotless.FormatterFunc.apply(FormatterFunc.java:32)
	at com.diffplug.spotless.FormatterStepImpl$Standard.format(FormatterStepImpl.java:82)
	at com.diffplug.spotless.FormatterStep$Strict.format(FormatterStep.java:88)
	at com.diffplug.spotless.Formatter.compute(Formatter.java:246)
	at com.diffplug.spotless.PaddedCell.calculateDirtyState(PaddedCell.java:203)
	at com.diffplug.gradle.spotless.IdeHook.performHook(IdeHook.java:58)
	at com.diffplug.gradle.spotless.SpotlessExtensionImpl.lambda$createFormatTasks$9(SpotlessExtensionImpl.java:86)
```

---

According to the context receivers proposal [here](https://github.com/Kotlin/KEEP/blob/72b23482acc576745af5f3f6ccbcc232367793b7/proposals/context-receivers.md?plain=1#L7), the earliest Kotlin release that supports them is `1.6.20-M1`. However, ktfmt seems to be using `1.6.10` as seen [here](https://github.com/facebook/ktfmt/blob/8bde75c826b84ae027955e6a5f3ebe671abf09c9/core/pom.xml#L19). Am I correct to assume that that's the issue?

---

#314 seems related, but according to the comments in that issue, ktfmt at one point worked when using context receivers. This issue here instead explains that ktfmt is basically crashing due to an issue parsing the source code.

bddckr:
Resolves #397, #314 and #374.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -95,6 +95,7 @@ import org.jetbrains.kotlin.psi.KtQualifiedExpression
 import org.jetbrains.kotlin.psi.KtReferenceExpression
 import org.jetbrains.kotlin.psi.KtReturnExpression
 import org.jetbrains.kotlin.psi.KtScript
+import org.jetbrains.kotlin.psi.KtScriptInitializer
 import org.jetbrains.kotlin.psi.KtSecondaryConstructor
 import org.jetbrains.kotlin.psi.KtSimpleNameExpression
 import org.jetbrains.kotlin.psi.KtStringTemplateExpression
@@ -2450,6 +2451,7 @@ class KotlinInputAstVisitor(
   override fun visitScript(script: KtScript) {
     markForPartialFormat()
     var lastChildHadBlankLineBefore = false
+    var lastChildIsContextReceiver = false
     var first = true
     for (child in script.blockExpression.children) {
       if (child.text.isBlank()) {
@@ -2459,6 +2461,8 @@ class KotlinInputAstVisitor(
       val childGetsBlankLineBefore = child !is KtProperty
       if (first) {
         builder.blankLineWanted(OpsBuilder.BlankLineWanted.PRESERVE)
+      } else if (lastChildIsContextReceiver) {
+        builder.blankLineWanted(OpsBuilder.BlankLineWanted.NO)
       } else if (child !is PsiComment &&
           (childGetsBlankLineBefore || lastChildHadBlankLineBefore)) {
         builder.blankLineWanted(OpsBuilder.BlankLineWanted.YES)
@@ -2466,6 +2470,7 @@ class KotlinInputAstVisitor(
       visit(child)
       builder.guessToken(";")
       lastChildHadBlankLineBefore = childGetsBlankLineBefore
+      lastChildIsContextReceiver = child is KtScriptInitializer && child.firstChild?.firstChild?.firstChild?.text == "context"
       first = false
     }
     markForPartialFormat()
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -1469,8 +1469,12 @@ class KotlinInputAstVisitor(
 
   override fun visitClassOrObject(classOrObject: KtClassOrObject) {
     builder.sync(classOrObject)
+    val contextReceiverList = classOrObject.getStubOrPsiChild(KtStubElementTypes.CONTEXT_RECEIVER_LIST)
     val modifierList = classOrObject.modifierList
     builder.block(ZERO) {
+      if (contextReceiverList != null) {
+        visitContextReceiverList(contextReceiverList)
+      }
       if (modifierList != null) {
         visitModifierList(modifierList)
       }
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -1661,6 +1661,7 @@ class KotlinInputAstVisitor(
             prefix = "(",
             postfix = ")",
             breakAfterPrefix = false,
+            breakBeforePostfix = false
     )
     builder.forcedBreak()
   }
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -52,6 +52,7 @@ import org.jetbrains.kotlin.psi.KtCollectionLiteralExpression
 import org.jetbrains.kotlin.psi.KtConstantExpression
 import org.jetbrains.kotlin.psi.KtConstructorDelegationCall
 import org.jetbrains.kotlin.psi.KtContainerNode
+import org.jetbrains.kotlin.psi.KtContextReceiverList
 import org.jetbrains.kotlin.psi.KtContinueExpression
 import org.jetbrains.kotlin.psi.KtDelegatedSuperTypeEntry
 import org.jetbrains.kotlin.psi.KtDestructuringDeclaration
@@ -124,6 +125,7 @@ import org.jetbrains.kotlin.psi.psiUtil.children
 import org.jetbrains.kotlin.psi.psiUtil.getPrevSiblingIgnoringWhitespace
 import org.jetbrains.kotlin.psi.psiUtil.startOffset
 import org.jetbrains.kotlin.psi.psiUtil.startsWithComment
+import org.jetbrains.kotlin.psi.stubs.elements.KtStubElementTypes
 
 /** An AST visitor that builds a stream of {@link Op}s to format. */
 class KotlinInputAstVisitor(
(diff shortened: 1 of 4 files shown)


bddckr:
I mostly have no idea what I'm doing 😅, but the added tests pass with these changes.

Please let me know if I'm missing anything. In particular, I've not been able to find how to lint my changes, [as requested by the contributing docs](https://github.com/facebook/ktfmt/blob/19363984e8f2c765547daacb568eec6e2b608919/CONTRIBUTING.md?plain=1#L14).

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -1626,6 +1635,16 @@ class KotlinInputAstVisitor(
     builder.forcedBreak()
   }
 
+  /** Example `context(Logger, Raise<Error>)` */
+  override fun visitContextReceiverList(contextReceiverList: KtContextReceiverList) {
+    builder.sync(contextReceiverList)
+    builder.token("context")
+    builder.token("(")
+    visitEachCommaSeparated(contextReceiverList.contextReceivers())
```
nreid260, line 1648: visitEachCommaSeparated has parameters for enclosing parens, to handle some weird indentation cases.

core/src/test/java/com/facebook/ktfmt/format/TokenizerTest.kt
```diff
@@ -110,4 +110,25 @@ class TokenizerTest {
         .containsExactly(0, -1, 1, 2, 3, -1, 4, -1, 5, 6, 7)
         .inOrder()
   }
+
+  @Test
+  fun `Context receivers are parsed correctly`() {
+    val code = """
+      |class A {
+      |  context(Logger, Raise<Error>)
```
nreid260, line 119: What happens if the list of receivers is longer than the max line length?

What happens if there's a trailing comma?

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
bddckr: Pushed a commit to use that, thank you for pointing that out!

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -1461,8 +1470,12 @@ class KotlinInputAstVisitor(
 
   override fun visitClassOrObject(classOrObject: KtClassOrObject) {
     builder.sync(classOrObject)
+    val contextReceiverList = classOrObject.getStubOrPsiChild(KtStubElementTypes.CONTEXT_RECEIVER_LIST)
```
nreid260, line 1473: There's a method for this: getContextReceiverList()

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -1638,6 +1652,20 @@ class KotlinInputAstVisitor(
...
 
+  /** Example `context(Logger, Raise<Error>)` */
+  override fun visitContextReceiverList(contextReceiverList: KtContextReceiverList) {
+    builder.sync(contextReceiverList)
+    builder.token("context")
+    visitEachCommaSeparated(
+            contextReceiverList.contextReceivers(),
+            prefix = "(",
+            postfix = ")",
+            breakAfterPrefix = false,
```
nreid260, line 1663: You probably don't want to specify these. They override the defaults to say "never break before or after the prefix, regardless of line length".

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -6674,6 +6674,48 @@ class FormatterTest {
...
+
+    val expected =
+        """
+      |context(Something)
+      |class A {
+      |  context(
+      |  // Test comment.
+      |  Logger,
+      |  Raise<Error>)
+      |  @SomeAnnotation
```
nreid260, line 6707: If the context receivers break to multiple lines, I'd expect the list of annotations to also break. The repo owners may disagree though.

Either way, there should be a test for the case of multiple annotations.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
nreid260, line 1: I may have missed it, but there don't seem to be changes for context receivers on lambda parameters:

```
fun foo(cb: context(String) (Int) -> Unit) {
  cb("", 0)
}
```

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -6674,6 +6674,48 @@ class FormatterTest {
...
+            .trimMargin()
+
+    val expected =
+        """
+      |context(Something)
+      |class A {
+      |  context(
+      |  // Test comment.
+      |  Logger,
+      |  Raise<Error>)
```
nreid260, line 6706: This is very subtle, but I'm specifically curious what happens when there's a comment between the last receiver type and the paren. That positioning is often buggy.

core/src/test/java/com/facebook/ktfmt/format/TokenizerTest.kt
nreid260: Thanks. It's very weird to me that trailing commas aren't accepted in the context receiver list. I suspect that's a bug in kotlinc, but nothing we can do right now.

JavierSegoviaCordoba:
Even the `context` syntax will have more problems, currently, the main issue is the space generated between `context` and the function itself. I have not ever seen issues when it is combined with annotations, or when it is used on lambdas (`fun foo(block: context(Foo, Bar) () -> Unit)` or adding basic comments to them.

I haven't seen any crashes too which could be a real problem.

Can we get a quick fix for the main issue and improve it later by merging this PR soon?

Even Kotlin itself has issues that need to be fixed, for example, trailing commas are not working, annotation position is affecting and it will not affect in the future (I think), and so on. So spending a lot of time when maybe in the future the feature is changed a bit cannot be productive too.

I use ktfmt even in playground projects without any other plugin than the IDE one, and it is really annoying to fix those issues manually as pressing `cmd + alt + L` is automatic.

core/src/test/java/com/facebook/ktfmt/format/TokenizerTest.kt
```diff
@@ -110,4 +110,28 @@ class TokenizerTest {
...
+
+    val file = Parser.parse(code)
+    val tokenizer = Tokenizer(code, file)
+    file.accept(tokenizer)
+
+    assertThat(tokenizer.toks.map { it.originalText })
+            .containsExactly("context", "(", "Something", ")", "\n", "class", " ", "A", " ", "{", "\n", "  ", "context", "(", "\n", "  ", "// Test comment.", "\n", "  ", "Logger", ",", " ", "Raise", "<", "Error", ">", ")", "\n", "  ", "fun", " ", "test", "(", ")", " ", "{", "}", "\n", "}")
+            .inOrder()
+    assertThat(tokenizer.toks.map { it.index })
+            .containsExactly(0, 1, 2, 3, -1, 4, -1, 5, -1, 6, -1, -1, 7, 8, -1, -1, 9, -1, -1, 10, 11, -1, 12, 13, 14, 15, 16, -1, -1, 17, -1, 18, 19, 20, -1, 21, 22, -1, 23)
```
davidtorosyan, line 134: nit: not a problem for this diff, but this unit test file isn't very maintainable. We should add some helpers to make it so that it's clear what's actually being tested, and how the numbers relate to the strings.

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -6674,6 +6674,48 @@ class FormatterTest {
...
+      |}
+      |"""
+            .trimMargin()
+
+    val expected =
+        """
+      |context(Something)
+      |class A {
+      |  context(
+      |  // Test comment.
```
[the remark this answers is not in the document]
davidtorosyan: +1 to this and your other comments in this test. Don't think any of it is blocking though.

bddckr:
Thanks for merging this as-is for now!

Sadly, I've not had time yet to come back to this to clean up and split the tests, add more test cases based on the raised PR review comments, and handle more scenarios. Depending on how soon someone else might get to that, I'm still interested in contributing further improvements. However, I may not get to that before the end of this year.

PR #401, opened 2023-05-11
Don't insert blank lines between line comments at the end of files

nreid260:
There was already code/testing for this, but it wasn't covering a strange case where a comment was the first line of the file.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -2390,16 +2390,19 @@ class KotlinInputAstVisitor(
 
   override fun visitKtFile(file: KtFile) {
     markForPartialFormat()
-    var importListEmpty = false
+    val importListEmpty = file.importList?.text?.isBlank() ?: true
+
     var isFirst = true
     for (child in file.children) {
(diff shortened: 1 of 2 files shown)


nreid260:
@cgrushko

cgrushko:
heh, we didn't feel this internally because we have a blank line between the first line (comment) and the package statement :) 
Can you add a test that makes sure that an existing blank line is preserved?

nreid260:
> heh, we didn't feel this internally because we have a blank line between the first line (comment) and the package statement :) Can you add a test that makes sure that an existing blank line is preserved?

Done

cgrushko:
Apologies for the delay, everybody seems to be unavailable this week at the same time :(

nreid260:
Yeah, I know the feeling. Here too.

hick209:
Overall this looks like a good improvement.
In our tests to see what changes though, we found this regression.

With this change this code here, which should not be touched
```
// License

package /* comment */ com.example.ktfmt /* comment */

import com.example.ImportOne
import com.example.ImportTwo
// ...
```

Gets formatted to this – removing empty line separating package declaration and imports
```
// License

package /* comment */ com.example.ktfmt /* comment */
import com.example.ImportOne
import com.example.ImportTwo
// ...
```

Could you take a look into this before we proceed with the merging?

PR #402, opened 2023-05-14
remove newlines first after block.

jiqimaogou:


nreid260:
Please add or modify a test for this change. It makes it easier to understand the effect, and also prevents us from reverting the new behaviour.

hick209:
@jiqimaogou I've tested the change locally in order to find out what it does.

For context, basically makes changes like this:
```
class Foo {

  fun foo() {

    // foo content
  }
  ...
}
```
to
```
class Foo {
  fun foo() {
    // foo content
  }
  ...
}
```
Removing the space after the bracket open.

While I don't think this by itself is a bad change, I believe it would require a discussion to happen before we implement this for everyone.
At the very least this should be an optional thing.


@jiqimaogou as this is an old PR and you have not replied to the comment by @nreid260 on May (4 months ago) I'll close it for now, but please feel free to reopen with changes and/or start the discussion around this.

PR #414, opened 2023-08-17
Keep imports from the same package if the name is overloaded

nreid260:


nreid260:
@cgrushko

core/src/main/java/com/facebook/ktfmt/format/RedundantImportDetector.kt
```diff
@@ -85,7 +85,7 @@ internal class RedundantImportDetector(val enabled: Boolean) {
 
   private var thisPackage: FqName? = null
 
-  private val usedReferences = OPERATORS.toMutableSet()
+  private val usedReferences: MutableSet<String> = OPERATORS.toMutableSet()
```
hick209, line 88: nit: Is the explicit type here really necessary?

core/src/main/java/com/facebook/ktfmt/format/RedundantImportDetector.kt
```diff
@@ -160,20 +159,23 @@ internal class RedundantImportDetector(val enabled: Boolean) {
...
+        .mapValues { it.value.size }
+
+    return importCleanUpCandidates
+      .filter{
+        val isUsed = it.identifier in usedReferences
+        val isFromThisPackage = it.importedFqName?.parent() == thisPackage
+        val hasAlias = it.alias != null
+        val isOverload = requireNotNull(identifierCounts[it.identifier]) > 1
+        // Remove if...
+        !isUsed || (isFromThisPackage && !hasAlias && !isOverload)
```
hick209, line 174: We might actually want to pivot this to just remove if it's unused, which keeping things from the same package always.

We have internal reasons to want this change, but this is something we likely will do in the future and if you state that this is not something you'd want we could put it under a config. Please let us know.

Things are fine for this PR as they are though, just to be clear

hick209 (approved):

core/src/main/java/com/facebook/ktfmt/format/RedundantImportDetector.kt
nreid260: I do generally prefer the behaviour of removing imports for the local package, since it keeps things consistent. Otherwise, some local package elements will be imported, and others won't be, depending on the whims of various authors.

The case I'm trying to fix has to do with overload resolution for Truth. The issue seems to be that importing `Truth.assertThat` causes kotlinc to always use an overload from the `Truth` class, even when there's a local package function named `assertThat`, which has narrower typing. Part of the problem here is there's a definition `Truth.assertThat(x: Object)` which matches ~everything, which is a bad but entrenched design.

core/src/main/java/com/facebook/ktfmt/format/RedundantImportDetector.kt
nreid260: Up to you

core/src/main/java/com/facebook/ktfmt/format/RedundantImportDetector.kt
```diff
@@ -112,11 +112,10 @@ internal class RedundantImportDetector(val enabled: Boolean) {
     importCleanUpCandidates =
         importList.imports
             .filter { import ->
+              val identifier = import.identifier ?: return@filter false
               import.isValidImport &&
-                  !import.isAllUnder &&
```
cgrushko, line 116: why remove the isAllUnder?

core/src/main/java/com/facebook/ktfmt/format/RedundantImportDetector.kt
cgrushko: let's drop for now to reduce blame

core/src/main/java/com/facebook/ktfmt/format/RedundantImportDetector.kt
nreid260: isAllUnder implies identifier == null, which is already considered.

hick209 (approved):

PR #416, opened 2023-08-24
Fix double indentation in Elvis chains

nreid260:
This is consistent with other binary operators, and prevents a strange second indent in Elvis chains.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -1207,20 +1207,31 @@ class KotlinInputAstVisitor(
     val leftMostExpression = parts.first()
     visit(leftMostExpression.left)
     for (leftExpression in parts) {
-      when (leftExpression.operationToken) {
-        KtTokens.RANGE -> {}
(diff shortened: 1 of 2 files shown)


nreid260:
@cgrushko 

I feel this might be a contentious change. What I'm mostly hoping to fix is this output:
```
-------------------------------
      foo =
        option1()
          ?: option2() ?: option3()
            ?: option4() ?: option5()
            ?: error()
```

hick209:
What were your expectations for that particular use case there?

Would this do?
```
----------------------------------------
      foo =
        option1()
          ?: option2() ?: option3()
          ?: option4() ?: option5()
          ?: error()
```

nreid260:
That proposed formatting is also acceptable to me. Although UNIFIED breaks might be preferable since the operators all have the same precedence:

---------------------------------------------------
      foo =
        option1()
          ?: option2()
          ?: option3()
          ?: option4()
          ?: option5()
          ?: error()

nreid260:
@hick209 @cgrushko @davidtorosyan Is there a strong opinion here? I'd like to move forward in some direction

nreid260:
I took another look and it was surprisingly easy to get the intended left-alignment. Hopefully that moves this from a decision to a bug fix.

hick209:
No strong opinions from my end other than "this looks like a nice improvement to me"

nreid260:
Great. I believe the latest version is uncontroversial as well.

PR #417, opened 2023-09-13
Use inExpression in a nullsafe way

nreid260:
peekLast is a JDK method using platform types for the return value.


nreid260:
@hick209

hick209:
Is this fixing an issue that you have seem or just a precaution?

nreid260:
I mostly care because our internal build has some additional null checking for JDK classes, which is triggering on this code. However, this version will also crash on unbalanced use of the stack, which is nice.

hick209:
Ran into merge conflicts while trying to land this. @nreid260 would you mind rebasing it?

hick209:
Nevermind, it was simple enough. I've rebased it

Issue #388, opened 2023-02-28
Apparent inconsistencies in which lines are or aren't reformatted within a statement


kevinb9n:
(This may be IDEA-plugin-specific, not sure.)

Here's a fun bug I had

```kotlin
  val allNodes: Set<Node> by lazy {
    setOf<Node>() +
        id +
        dependencies.map { it.typeExpr } +
        effectsRaw +
        defaults.allNodes  // only this line is modified
        extraNodes
  }
```

Note that if the formatter had reformatted this code properly I'd have caught the bug much more quickly than I did.

Here's what I'm seeing:

1. Hit "reformat code" keystroke in IDEA editor (with nothing selected): **does not** fix
2. Same but select *any character* in the file first: **does** fix
3. Right-click filename in Project pane, "reformat code", check "VCS only": **does not** fix
4. Right-click filename in Project pane, "reformat code", don't check "VCS only": **does** fix

What I would have expected is something like:

1. does fix (I assume the keystroke with nothing selected would mean the whole file or at least all changed lines)
2. fixes only if the selection includes the affected area
3. does fix
4. does fix

I also tried making a simpler kind of style error (an extra space before a paren) elsewhere in the file, and it was fixed in all four cases (even when the selected code was nowhere near the problem).

davidtorosyan:
Hey @kevinb9n, thanks for reporting.

Before talking about the specific issues, I want to make sure we're talking about the same thing.

When I format with the [online tool](https://facebook.github.io/ktfmt/) (using java-like), I get this:
```kt
  val allNodes: Set<Node> by lazy {
    setOf<Node>() +
        id +
        dependencies.map { it.typeExpr } +
        effectsRaw +
        defaults.allNodes // only this line is modified
    extraNodes
  }
```

This is what you mean by "**does** fix", right?

---

I don't know a lot about the IDEA plugin, but my guess is that it only bothers reformatting if either (a) the file has changed, (b) you select a piece of code to format.

If that's true, then:
* (1) is expected (because nothing has changed)
* (2) is expected because you're explicitly formatting some code, and ktfmt always formats the whole file

I don't know what the VCS option means for (3) and (4).

cgrushko:
I think the IDEA plugin is just old - I had trouble publishing the latest version due to some old APIs we're using.

kevinb9n:
@davidtorosyan Correct, that is fixed.

[...]

PR #420, opened 2023-09-18
Update ktfmt plugin for IntelliJ IDEA

### LINKED ISSUE #399: IDEA plugin is not published since 0.43
There's nothing much to add to the title: there are no nor 0.43 nor 0.44 version of IDEA plugin present on JetBrains marketplace

greyhairredbear:
Changes: 
- update build dependencies
  - gradle wrapper
  - intellij gradle plugin
- use `AsyncDocumentFormattingService` implementation to format code with `ktfmt` (adjusted from `google-java-format` IDEA plugin)

~~Intended to fix #399~~ (IDEA plugin seems to have been published for `0.46`)


hick209:
I believe this will also fix #388

greyhairredbear:
Note for reviewers: The `google-java-format` plugin contains a test for `GoogleJavaFormatFormattingService`, which I haven't adjusted and ported for `ktfmt`. If required however, writing a similar test for `KtfmtFormattingService` should be possible.

greyhairredbear:
@cgrushko @strulovich Seems running the workflows on this PR need approval from a maintainer - would it be possible to run the workflows? 

Other than that, is there a way to get feedback on the changes? Could I help in any way?

hick209:
Sorry the super long delay here @greyhairredbear. I went on paternity leave in September (son born on 22 of Sept) and once I was back early this year there were many things pilled up that took me a while to  go through.

I'm importing this internally for review and we should have some feedback for you this week.

greyhairredbear:
> Sorry the super long delay here @greyhairredbear. I went on paternity leave in September (son born on 22 of Sept) and once I was back early this year there were many things pilled up that took me a while to go through.
> 
> I'm importing this internally for review and we should have some feedback for you this week.

Hi, no need for excuses here. Congrats on your son, being a father of three, I completely understand the priorities here :) 

Happy about any feedback or change requests

ktfmt_idea_plugin/src/main/java/com/facebook/ktfmt/intellij/KtfmtFormattingService.java
```diff
@@ -0,0 +1,97 @@
...
+            } catch (FormatterException e) {
+                request.onError(
+                        Notifications.PARSING_ERROR_TITLE,
+                        Notifications.parsingErrorMessage(request.getContext().getContainingFile().getName()));
+            }
+        }
+
+        @Override
+        public boolean isRunUnderProgress() {
+            return true;
```
hick209, line 89: Any particular reason not to return a boolean flag here? We could set its value in the `run()` method

hick209 (commented):
I just have one question here, since I'm just mildly familiar with IJ API

hick209 (changes requested):
We are good to ship this internally, but I'm sending back to you to get an answer from the question I've asked above.

ktfmt_idea_plugin/src/main/java/com/facebook/ktfmt/intellij/KtfmtFormattingService.java
greyhairredbear: Sorry it took me so long to respond - I was quite busy at work and had to look that up. Unfortunately, `isRunUnderProgress` is a bit of an ambiguous name. 

I quote from the documentation: 
> Returns:
True if the task must be run under progress (a progress indicator is created automatically). Otherwise, the task is responsible for visualizing the progress by itself, it is just started on a background thread.

So the return value here is not about if the formatter is currently running, but rather if IntelliJ should handle showing a progress indicator to users.

hick209 (approved):

PR #434, opened 2023-12-16
Preserve semicolons in enums with members but no entries

nreid260:
Inspired by https://github.com/facebook/ktfmt/pull/425


nreid260:
@strulovich

nreid260:
@davidtorosyan @strulovich

Issue #201, opened 2021-03-28
Support JDK 8


cgrushko:
Currently, ktfmt requires JDK 11 or newer because its dependency, google-Java-format 1.8, uses it in its Java fronted. 

This minimum requirement is problematic for ktfmt because:
1. The plug-in doesn't work on Android Studio 4.1.1 without manually running AS under a non-standard JRE. 
2. Any open-source project that uses ktfmt must also use JDK 11 or newer. 

Solution: for now, we'll fork GJF's JavaOutput (rename it to KotlinOutput) and downgrade our dependency to GJF 1.7, which supported JDK 8. 

In the future, if there are considerable changes to the GJF language agnostic parts that we want, we can do either of
1. Convince GJF maintainers to cut a version that only contains the engine (nothing there inherently needs a new JDK), or
2. Fork the GJF engine and bundle it into ktfmt.

cgrushko:
1. ktfmt itself is easy to fix
2. ktfmt_idea_plugin is problematic: if I build against IJ 203, I get a version mismatch, because the 203 jars are built using JDK 11. I f I build against IJ 201, ktfmt fails because it no longer can handle Kotlin 1.3 on the classpath. Which leads to the next point -
3. The ktfmt plugin runs inside IJ with IJ's Kotlin stdlib on the classpath, which is not necessarily what ktfmt was built and intended to run with. That's very meh. The straightforward solution is to bundle ktfmt inside the plugin and load it using a separate class loader.

JavierSegoviaCordoba:
4.2 RC is here, and Canary and Compose needs JDK 11 I think. Do you think it is better to support 1.8 instead of wait? Probable 4.2 stable is going to be presented in Google IO in May

JavierSegoviaCordoba:
4.2 is released and it uses JDK 11 by default 🎉

JavierSegoviaCordoba:
I am using ktfmt in arrow-kt repos and we need to use java 8, do you think you can fix this some day?

hick209:
@JavierSegoviaCordoba, is this request for Java 8 support still current?

JavierSegoviaCordoba:
No, you can close this

PR #452, opened 2024-04-02
Update google-java-format to 1.22.0

nreid260:
This update adds support for multiline tokens, which fixes several bugs related to multiline strings.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -179,7 +179,7 @@
         <dependency>
             <groupId>com.google.googlejavaformat</groupId>
             <artifactId>google-java-format</artifactId>
-            <version>1.8</version>
+            <version>1.22.0</version>
         </dependency>
         <dependency>
             <groupId>junit</groupId>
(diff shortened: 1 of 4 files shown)


hick209:
I don't know if we could actually land this as internally we're barely at 1.21-ish (more like 1.19 really), but I am currently working on changing that.

Let me try pulling it and see what happens

hick209:
It fails the tests internally. Looks like this would have to wait until we have performed the upgrade on our end.

Also does this clash with this issue here? https://github.com/facebook/ktfmt/issues/201
I'm no Java version number and flavors and all expert here, so I'm always confused on that area

nreid260:
Within Google, google-java-format is built using JDK11, so I presume that's why the Marven artifact specifies it. I don't think there are currently any Java 11 features being used, but they may slip in as time passes.

hick209:
In the meantime do you want to change this to 1.21?

nreid260:
> In the meantime do you want to change this to 1.21?

The multiline fix I care about is in 1.22.0

nreid260:
😀

PR #456, opened 2024-04-15
Remove duplicated line in editorconfig for dropbox style

sabufung30:
Remove `ij_java_names_count_to_use_import_on_demand = 9999` in `.editorconfig-dropbox`


hick209:
@sabufung30 thank you so much for the contribution!

Before we move forward, could you please add more context on why this change is required?

hick209:
Oof, sorry I missed that. Thanks for the fix 😃

PR #458, opened 2024-04-23
Stop emitting multiple trailing commas in enum entry lists with multiline items

nreid260:


nreid260:
@hick209

hick209:
Sure, I'm already preparing a release now, but we can do one right afterwards once we land this fix

nreid260:
Hurray!

PR #469, opened 2024-05-21
chore: update google style editorconfig

chenrui333:


hick209:
@chenrui333 would you mind adding context on what this change affects/does?

docs/editorconfig/.editorconfig-google
```diff
@@ -6,42 +6,38 @@
...
 
-[{*.kt,*.kts}]
-indent_style = space
-insert_final_newline = true
-max_line_length = 100
-indent_size = 2
-ij_continuation_indent_size = 2
-ij_java_names_count_to_use_import_on_demand = 9999
+[{*.gradle.kts,*.kt,*.kts,*.main.kts}]
+indent_size = 4
```
nreid260, line 10: --google-style indents everything by 2 spaces at a time, so this is almost certainly inconsistent.

nreid260:
I skimmed it, but I don't use IntelliJ, and I'm not sure anyone here that does is using an editorconfig. We enforce that they run ktfmt instead.

chenrui333:
this was a dump from the intellij config

docs/editorconfig/.editorconfig-google
chenrui333: though there is no official style guide in https://github.com/google/styleguide, this is the closest one by google team.
https://developer.android.com/kotlin/style-guide#whitespace

And this is the indentation guide
image

chenrui333: happy to change to indent_size to 2 though.

chenrui333: @nreid260 correct me if I miss anything.

nreid260: Our internal style guide is based on the public Android one you linked to, but with substantial modifications. I don't know if there are plans to publish it, but --google-style exists to enforce it. I admit things could be better.

To be clear, I have no stake in this editorconfig file. As far as I'm concerned, Google users will run ktfmt and accept its output. We have no need for IntelliJ to replicate the style.

hick209:
@chenrui333, any particular problem you are trying to solve here?

Since there's not much information on the PR at the moment I'll close it, but feel free to reopen if you believe this is something we should merge.

PR #467, opened 2024-05-17
Refactor CLI argument parsing

grodin:
Working towards #391. 

I wanted to open this to get feedback on the approach.

I could have chosen to use exceptions, but I find sealed classes much more composable and I already anticipate adding another class to the `ParseResult` hierarchy to represent a situation where a message needs to be shown to the user but isn't an error (`--help` is the obvious example). 

Closes #465 more or less as a side-effect.

PR #470, opened 2024-05-26
feat(cli): support max width arg

awe123343:
I have a project wanna moving from ktfmt in spotless gradle plugin to cli, but have a blocker right now due to we set max width to 120 with ktfmt spotless gradle plugin. Hence supporting this in CLI, looks like we support customising this in both spotless plugin and web UI, but just not CLI.


hick209 (changes requested):
We're discussing internally whether we want these to be available on the command line.

BTW there are merge conflicts now that we landed #467

awe123343:
> We're discussing internally whether we want these to be available on the command line.
> 
> BTW there are merge conflicts now that we landed #467

Rebased and solved conflicts while waiting for discussion

hick209:
This would go [against this principle here](https://github.com/google/google-java-format/wiki/FAQ#i-just-need-to-configure-it-a-bit-differently-how) so we overall don't want to support it.

I understand this is likely not the answer you wanted, but luckily there's an easy workaround.
You can build a simple Java program and runs ktfmt for you (from source) where you set it up appropriately as you can control the max width param from there.

I hope that is helpful to your use case and regardless I'm thankful for contributions to ktfmt

Issue #472, opened 2024-06-04
Google style with trailing comma does not maxWidth correctly


Rawa:
Hello!

There seems to be an off-by-one error when using Google-style and trailing commas. I just tried out the Google Style with trailing commas and saw the following scenario:
```
    @Test
    fun testShowBillingErrorPaymentButton() =
        composeExtension.use {
            // Arrange
            setContentWithTheme {
                AccountScreen(
                    state =
                        AccountUiState.default()
                            .copy(billingPaymentState = PaymentState.Error.Billing),
                    uiSideEffect = MutableSharedFlow<AccountViewModel.UiSideEffect>().asSharedFlow(),
                )
            }

            // Assert
            onNodeWithText("Add 30 days time").assertExists()
        }

```
The line:
```                    uiSideEffect = MutableSharedFlow<AccountViewModel.UiSideEffect>().asSharedFlow(),```
ends up being 101 characters even though I set maxWidth to 100.

If i change the named parameter to `uiSideEffects`, we see the following behavior, where it correctly wraps the line:
```
    @Test
    fun testShowBillingErrorPaymentButton() =
        composeExtension.use {
            // Arrange
            setContentWithTheme {
                AccountScreen(
                    state =
                        AccountUiState.default()
                            .copy(billingPaymentState = PaymentState.Error.Billing),
                    uiSideEffects =
                        MutableSharedFlow<AccountViewModel.UiSideEffect>().asSharedFlow(),
                )
            }

            // Assert
            onNodeWithText("Add 30 days time").assertExists()
        }
```

My configuration is as follows:
```
    configure<com.ncorti.ktfmt.gradle.KtfmtExtension> {
        googleStyle()
        blockIndent.set(4)
        continuationIndent.set(4)
        maxWidth.set(100)
        removeUnusedImports.set(true)
    }
```

nreid260:
The 1.0 issue mentioned merging trailing-comma management into all styles. If we do that, I think this issue is unblocked.

We could switch the default behaviour of trailing-commas to do "nothing", rather than today where they trigger a line-break. Once that's done, management becomes:
  - Insert trailing-commas everywhere
  - Pretty print
  - Remove unnecessary trailing-commas

Since the insertion happens before pretty-print, all the elements are already at their maximum length. We might get some lines that are unnecessarily wrapped after removal, but that's much less noticeable than exceeding max-width.

nreid260:
[...]

PR #473, opened 2024-06-04
Don't insert trailing commas over the max width

nreid260:
Closes https://github.com/facebook/ktfmt/issues/472

### DIFF
--- a/core/src/test/java/com/facebook/ktfmt/format/GoogleStyleFormatterKtTest.kt
+++ b/core/src/test/java/com/facebook/ktfmt/format/GoogleStyleFormatterKtTest.kt
@@ -1832,6 +1832,30 @@ class GoogleStyleFormatterKtTest {
     assertThatFormatting(code).withOptions(Formatter.GOOGLE_FORMAT).isEqualTo(expected)
   }
 
+  @Test
+  fun `trailing commas are not inserted over the max width`() =
+      assertFormatted(
+          """
+      |//////////////////
+      |fun foo() {
+      |  bar(
+      |    someArg,
+      |    justBeforeMax,
+      |  )
+      |  bar(
+      |    someArg,
+      |    exactlyAtMaaax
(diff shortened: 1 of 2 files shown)


nreid260:
@hick209

core/src/test/java/com/facebook/ktfmt/format/GoogleStyleFormatterKtTest.kt
```diff
@@ -1832,6 +1832,30 @@ class GoogleStyleFormatterKtTest {
...
+          """
+      |//////////////////
+      |fun foo() {
+      |  bar(
+      |    someArg,
+      |    justBeforeMax,
+      |  )
+      |  bar(
+      |    someArg,
+      |    exactlyAtMaaax
```
hick209, line 1847: Is that was @Rawa was referring to in #472?

What I understood was that the comma should have forced the line break, respecting the line max width option, which for this particular case here, where we cannot respect it, the comma would still be included.

@Rawa, is my understanding correct or were you also thinking something along the lines of what @nreid260 did here?

nreid260: This isn't an ideal solution, more of a hack to mitigate complaints. I can't think of an efficient algorithm for:
  - Remove unnecessary trailing commas
  - Insert line breaks between items in long lists
    - Insert trailing commas in broken lists

It's the last part that's being reported as a bug, because when the list item is exactly 100 columns, the trailing comma is inserted into column 101. Shorter items put the comma within bounds, longer items get wrapped during formatting.

Running prettyPrint twice would be slow, and also makes tests hard to debug.

hick209:
TBH I'd prefer to have the code as is vs this modification here.

I'm happy to hear arguments countering that though, but if you don't have any, just keep this PR closed.

Issue #391, opened 2023-03-06
Document @file CLI arguments


grodin:
Parsing args from a file if the first argument is `@filename` was introduced in [this commit][commit] but doesn't seem to be documented anywhere apart from a cryptic hint in cli help. I only found it because I got curious so went digging in the source, but it's actually a useful feature which deserves documenting!

[commit]: https://github.com/facebook/ktfmt/commit/0901f21f64bda7004788b5623ec23aef724972b1

hick209:
@grodin, that's a [common practice for CLIs](https://stackoverflow.com/a/13040192/1848826) and often [CLI tools offer that right out of the box](https://ajalt.github.io/clikt/advanced/#command-line-argument-files-argfiles), but sure we could document that.

Do you want to submit a PR for it?

grodin:
Can do. Should have some time next week.

grodin:
Starting work on this, I'd forgotten that the CLI is pretty barebones (e.g. lacks `--help` flag, only shows generic usage when failing to parse command line). Would you be interested in me improving the UX of the CLI a bit (adding `--help` mainly)?

Or should I just document the `@argfile` feature minimally, in the CLI and on the website?

Happy to do either, but might be overstepping the mark a bit with the first if you're wanting to keep the CLI ultra-minimal.

hick209:
I'd be happy to have it improved! 😃 

Just maybe make it into different PRs, if possible

grodin:
Yep, no problem

grodin:
How far can I go in rebuilding the CLI? 

One option is to go the whole hog and use [clikt][1] but obviously that's a new dependency and possibly bloating a pretty small Jar.

The current CLI behaviour is encoded in tests. Should I try to keep those tests passing or can I replace them?

[1]: https://github.com/ajalt/clikt

hick209:
Although I love clikt and use it in other projects, we intentionally didn't use it (or other cli libs) to keep the library slim.

Feel free to go as deep as you want on refactoring.
Once you have something, even if not really, share it so we can start the discussion in case we want to change somethings up

hick209:
Oh, just try not to change the CLI public API, unless it's really necessary. We're going with a 1.0 release, so that could warrant an API change.

grodin:
> Oh, just try not to change the CLI public API, unless it's really necessary. We're going with a 1.0 release, so that could warrant an API change.

[...]

PR #476, opened 2024-06-10
Further refactoring on CLI

grodin:
This doesn't add any features or fix any bugs, but does change the style option from `--dropbox-style` to `--style=dropbox` (and similarly for the other styles) as discussed in #391.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/cli/Main.kt
+++ b/core/src/main/java/com/facebook/ktfmt/cli/Main.kt
@@ -86,10 +86,7 @@ class Main(
       } catch (e: Exception) {
         1
       }
-    } else if (parsedArgs.stdinName != null) {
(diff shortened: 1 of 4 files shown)


grodin:
I've got a follow on PR ready to go which adds `--help`. I can happily push it into this branch but I thought I'd keep it separate to start with.

hick209:
As this one breaks the current CLI API contract, let's land it last as make it part of the 1.0.0 release.

Do you think we could land the `--help` one before this here?

grodin:
I've reverted the commit that adds `--style=` since the other stuff is just refactoring, and I'm quite keen to land that refactoring before I do too much more. I'll follow up immediately with `--help` and move `--style=` to a separate PR

hick209:
SGTM 😃

hick209 (approved):

PR #498, opened 2024-08-04
Treat KDoc `@property`s as special types of `@param`s, fixes #431

### LINKED ISSUE #431: KDoc: `@param` and `@property` sorted by type, not by position
When formatting the KDoc of a class, the `@param` and `@property` tags can be used to describe parameters. `@param` should be used for parameters that are not (visibly) stored (e.g. `foo` or `private val foo`), while property should be used for parameters that become visible properties (e.g. `val foo`). This is e.g. enforced by [detekt's OutdatedDocumentation rule](https://detekt.dev/docs/rules/comments/#outdateddocumentation).

Detekt requires these tags to be in the order of the primary constructor. Ktfmt insists on ordering these tags as `@param` (in constructor order), then `@property` (in constructor order). At the moment, it is impossible to satisfy both ktfmt and detekt.

### Expected by detekt

```kotlin
/**
 * @property foo …
 * @param bar …
 * @property baz …
 */
fun test(val foo: String, bar: String, val baz: String) {
    TODO()
}
```

### Formatted by ktfmt

```kotlin
/**
 * @param bar …
 * @property foo …
 * @property baz …
 */
fun test(val foo: String, bar: String, val baz: String) {
    TODO()
}
```

no-preserve-root:
Treat `@property` the same as `@param` (which matches the description in the KDoc documentation). This fixes an incompatibility between ktfmt and detekt's formatting rules.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/kdoc/ParagraphListBuilder.kt
+++ b/core/src/main/java/com/facebook/ktfmt/kdoc/ParagraphListBuilder.kt
@@ -571,18 +571,20 @@ class ParagraphListBuilder(
     return when {
       isPriority -> -1
       tag.startsWith("@param") -> 0
+      tag.startsWith("@property") -> 0
+      // @param and @property must be sorted by parameter order
+      // a @property is dedicated syntax for a main constructor @param that also sets a class property
       tag.startsWith("@return") -> 1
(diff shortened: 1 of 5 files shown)


hick209 (approved):

hick209:
We apologize for the delay. Thanks a ton for the contribution!

PR #501, opened 2024-08-07
Fix IntelliJ plugin on IJP 242

rock3r:
This PR fixes the issues in IJ 24.2, where the ktfmt crashes when being loaded and is essentially non-functional. It also removes a deprecated API usage in the same file.

Closes #494

PR #502, opened 2024-08-07
Convert IntelliJ plugin to Kotlin

rock3r:
> [!NOTE]
> This requires #501 to be merged first

This PR is the basis for future improvements of the plugin. In this PR, the plugin code is converted to Kotlin, the configurable is recreated with the Kotlin UI DSL, and the goal is to maintain the existing behaviours.

The only small change is that the style dropdown in the configurable is disabled when ktfmt is disabled.


hick209 (changes requested):
Overall this LGTM, but please mark the files as a move so we don't lose blame history.

rock3r:
Not sure how to do that? I didn't know you could override what diff thinks happened

rock3r:
@hick209 please advise so I can look into it tomorrow

rock3r:
@hick209 I tried looking into this but I did not find a way to force Git to consider files as moved. I had kept the actual conversion of the files to Kotlin to a dedicated commit, too, but that did not help. Also tried re-doing the conversion in another branch by committing the moving first, and the conversion second, but the conversion still results in pairs of adding/deleting

hick209:
Yeah I looked it up as well and noticed this is not really doable in git :-/ 

I'm used to working with Mercurial, which allows this with the following command:
```
hg mv FROM TO --after
```

No worries, I can try to do that internally once we pull your changes. I have to wait until we merge the other 2 PRs for this though.

Thanks a ton for you contributions!

ktfmt_idea_plugin/build.gradle.kts
```diff
@@ -15,48 +17,54 @@
...
+        untilBuild = provider { null }
+    }
+
+    publishing { token = System.getenv("JETBRAINS_MARKETPLACE_TOKEN") }
+
+    pluginVerification { ides { recommended() } }
 }
 
-spotless { java { googleJavaFormat("1.22.0") } }
+spotless { java { googleJavaFormat(libs.versions.googleJavaFormat.get()) } }
```
hick209, line 64: Now that we're on Kotlin, I guess we could even remove this

rock3r: True, will do. Btw, I formatted the Kotlin code with ktfmt, using the Meta style IIRC. Is that the correct one?

PR #511, opened 2024-08-20
fix: comment max width

rileymichael:
`FormattingOptions#maxWidth` is not respected for double slash comments. This is because it wasn't using the supplied value, and instead checked against the static property `com.google.googlejavaformat.java.Formatter.MAX_LINE_LENGTH`. I was unable to find any documentation / discussion that would indicate this as being intentional.

Test cases that deduced `maxWidth` actually relied on this behavior in some cases. Some of the fixes in 84dbfae449866787ba088d5ea9f0bad48b626cf5 simply adjusted the number of `/`, while others relied on a specific `maxWidth` and so I manually line-broke the comments to preserve the behavior under test.


hick209 (approved):

hick209:
Wow, great catch! Thanks for the fix 😃

Issue #495, opened 2024-07-17
Kotlin 2.0.20-Beta2: Exception when parsing a property accessor


ShikaSD:
Exception is thrown when parsing the following code:
```
class FilteringExecutor(
    private val delegate: ExecutorService = Executors.newSingleThreadExecutor()
) : Executor {
    var filterFunction: (Runnable) -> Boolean = { true }
        // failing expression
        set(value) {
            field = value
            reEnqueueDeferred()
        }
        // failing expression

    private fun reEnqueueDeferred()  {}
}
```

It appears that [`accessor.stub`](https://github.com/facebook/ktfmt/blob/main/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt#L1426) can be `null` with 2.0.20.

Full exception:
```
org.jetbrains.kotlin.com.intellij.psi.PsiInvalidElementAccessException: Element: class org.jetbrains.kotlin.psi.KtParameterList; stub hierarchy is invalid: org.jetbrains.kotlin.com.intellij.psi.impl.source.SubstrateRef$StubRef@23756ebe (class org.jetbrains.kotlin.com.intellij.psi.impl.source.SubstrateRef$StubRef) has null containing file stub
        at org.jetbrains.kotlin.com.intellij.psi.impl.source.SubstrateRef$StubRef.getContainingFile(SubstrateRef.java:110)
        at org.jetbrains.kotlin.com.intellij.extapi.psi.StubBasedPsiElementBase.getContainingFile(StubBasedPsiElementBase.java:235)
        at org.jetbrains.kotlin.com.intellij.extapi.psi.StubBasedPsiElementBase.getNode(StubBasedPsiElementBase.java:119)
        at org.jetbrains.kotlin.com.intellij.psi.PsiInvalidElementAccessException.getPsiInvalidationTrace(PsiInvalidElementAccessException.java:98)
        at org.jetbrains.kotlin.com.intellij.psi.PsiInvalidElementAccessException.<init>(PsiInvalidElementAccessException.java:61)
        at org.jetbrains.kotlin.com.intellij.psi.stubs.StubBase.<init>(StubBase.java:31)
        at org.jetbrains.kotlin.psi.stubs.impl.KotlinStubBaseImpl.<init>(KotlinStubBaseImpl.kt:34)
        at org.jetbrains.kotlin.psi.stubs.impl.KotlinPlaceHolderStubImpl.<init>(KotlinPlaceHolderStubImpl.java:28)
[...]

PR #513, opened 2024-08-27
Workaround non-null parent requirement of StubBase

jsjeon:


liutikas (commented):
Clever workaround!

jsjeon:
cc @hick209 as per https://github.com/facebook/ktfmt/issues/495 history

core/pom.xml
```diff
@@ -16,7 +16,7 @@
 
     <properties>
         <dokka.version>0.10.1</dokka.version>
-        <kotlin.version>1.8.22</kotlin.version>
+        <kotlin.version>2.0.20</kotlin.version>
```
hick209, line 19: If we don't touch this here would things still work?
If not, do you know what would be the min version we could put here?

hick209 (commented):
I really appreciate the code you put here, but I wonder if we need to update the Kotlin version.

My worry is that if we put Kotlin 2.0.20 here then maybe people trying to use ktfmt with an earlier version of Kotlin will have issues integrating with it.
I understand that running ktfmt from a binary should not be problem, but there are instances where it is used as a code integration, like in LSPs or running it as a deamon.

core/pom.xml
jsjeon: Not necessary to update the version; it was just handy to reproduce the issue. Back to the old version. :)

hick209:
nit: would you mind adding the change to the [`CHANGELOG.md`](https://github.com/facebook/ktfmt/blob/main/CHANGELOG.md)?

jsjeon:
> nit: would you mind adding the change to the [`CHANGELOG.md`](https://github.com/facebook/ktfmt/blob/main/CHANGELOG.md)?

Is it going to be Unreleased / fixed ?

hick209 (approved):

jsjeon:
Let me know what I need to do regarding Internal Linter error / test failures.

hick209:
@jsjeon it's just a minor thing, don't worry about it, I can take care of it 🙂

simonhauck:
Hi,

Would it be possible for you to create a release with this workaround? I encountered projects where it is now not possible to update the Kotlin version and have proper formatting checks, since this issue is also propagated in the [ktfmt-gralde](https://github.com/cortinico/ktfmt-gradle) plugin.

hick209:
Released on https://github.com/facebook/ktfmt/releases/v0.53

PR #520, opened 2024-11-07
Report an error on unclosed comments

nreid260:
Otherwise formatting succeeds and moves all subsequent code into the comment too. This is especially bad for nested comments, which have a surprising syntax.


nreid260:
@hick209

cortinico (approved):

PR #528, opened 2025-02-12
feat: Add more settings to CLI

kkoniuszy:
Exposes the FormattingOptions in the CLI. This brings it in line with the IJ plugin (#503) and the Gradle one

```
  --max-line-length=<length>        Max line length. Overrides the style.
  --block-indent=<size>             Size of the indent used when a new block is
                                        opened, in spaces. Overrides the style.
  --continuation-indent=<size>      Size of the indent used when a line is broken
                                        because it's too long, in spaces. Overrides
                                        the style.
  --manage-trailing-commas=(y|n)    Automatically remove and insert trailing commas.
                                        Overrides the style.
```


hick209:
Thanks for your effort in putting this PR together. Unfortunately we cannot accept it though.

We specifically do not want to make ktfmt more configurable via command line.

The rationale is the same as the one from [googlejavaformat here](https://github.com/google/google-java-format/wiki/FAQ#i-just-need-to-configure-it-a-bit-differently-how), on which ktfmt is based on.

PR #536, opened 2025-05-05
Support context parameters

### LINKED ISSUE #518: Missing support for context parameters on properties
Hello!

While introducing [context parameters](https://github.com/Kotlin/KEEP/blob/context-parameters/proposals/context-parameters.md) in my code, I noticed that ktfmt is not able to handle them on properties (at least properties in classes/interfaces, global scope seems to work for some reason).

Two possible errors are raised:

#### 0

```
context(Unit)
val someOuterProp: String get() = ""
```

This is fine.
I use `Unit` for simplicity's sake, but "real" interfaces exhibit the same behavior.

#### 1

Simple property with context parameter:

```kotlin
interface SomeInterface {
    context(Unit)
    val someInnerProp: String get() = ""
}
```

> error: expected token: 'context'; generated val instead

#### 2

```kotlin
interface BaseInterface {
    context(Unit)
    open val someInnerProp: String get() = ""
}

interface SomeInterface {
    context(Unit)
    override val someInnerProp: String get() = ""
}
```

Both of them can, in isolation, cause this:

> error: did not generate token "context"

---

A related issue I could find is #471, but that pertains to methods, not properties.
If I read the KEEP correctly, these cases should be legal, albeit experimental, Kotlin.

ZacSweers:
Resolves #518

Note that in order for `LightVirtualFile` to correctly parse this, it did require updating to the Kotlin 2.2.0 betas. Since this ships as a fat jar I think it would be fine to build against this in `main`, but up to y'all!

This adds a couple new tests for it + ensures backward compatibility with the previous context receivers syntax.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -22,7 +22,7 @@
 
   <properties>
     <dokka.version>0.10.1</dokka.version>
-    <kotlin.version>2.0.21</kotlin.version>
+    <kotlin.version>2.2.0-Beta2</kotlin.version>
     <kotlin.compiler.incremental>true</kotlin.compiler.incremental>
     <kotlin.compiler.jvmTarget>1.8</kotlin.compiler.jvmTarget>
     <main.class>com.facebook.ktfmt.cli.Main</main.class>
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -1689,12 +1689,19 @@ class KotlinInputAstVisitor(
     builder.forcedBreak()
   }
 
-  /** Example `context(Logger, Raise<Error>)` */
+  /**
+   * Example `context(logger: Logger, raise: Raise<Error>)`
+   *
+   * Note this also supports the legacy receiver format of `context(Logger, Raise<Error>)`
+   * for backward compatibility.
+   */
   override fun visitContextReceiverList(contextReceiverList: KtContextReceiverList) {
     builder.sync(contextReceiverList)
     builder.token("context")
+    val listToVisit =
+        contextReceiverList.contextParameters().ifEmpty { contextReceiverList.contextReceivers() }
     visitEachCommaSeparated(
-        contextReceiverList.contextReceivers(),
+        listToVisit,
         prefix = "(",
         postfix = ")",
         breakAfterPrefix = false,
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -1692,8 +1692,8 @@ class KotlinInputAstVisitor(
   /**
    * Example `context(logger: Logger, raise: Raise<Error>)`
    *
-   * Note this also supports the legacy receiver format of `context(Logger, Raise<Error>)`
-   * for backward compatibility.
+   * Note this also supports the legacy receiver format of `context(Logger, Raise<Error>)` for
+   * backward compatibility.
    */
   override fun visitContextReceiverList(contextReceiverList: KtContextReceiverList) {
     builder.sync(contextReceiverList)
--- a/core/src/main/java/com/facebook/ktfmt/format/Parser.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/Parser.kt
@@ -16,7 +16,6 @@
 
 package com.facebook.ktfmt.format
 
-import org.jetbrains.kotlin.cli.common.CLIConfigurationKeys
 import org.jetbrains.kotlin.cli.common.messages.MessageRenderer.PLAIN_RELATIVE_PATHS
 import org.jetbrains.kotlin.cli.common.messages.PrintingMessageCollector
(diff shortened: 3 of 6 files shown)


core/src/main/java/com/facebook/ktfmt/format/Parser.kt
```diff
@@ -51,7 +51,7 @@ object Parser {
     val disposable = Disposer.newDisposable()
     val configuration = CompilerConfiguration()
     configuration.put(
-        CLIConfigurationKeys.MESSAGE_COLLECTOR_KEY,
+        CommonConfigurationKeys.MESSAGE_COLLECTOR_KEY,
```
ZacSweers, line 54: This was opportunistic, previous constant was deprecated in favor of this one

core/src/test/java/com/facebook/ktfmt/format/TokenizerTest.kt
```diff
@@ -220,6 +220,129 @@ class TokenizerTest {
...
+      |class A {
+      |  context(
+      |  // Test comment.
+      |  logger: Logger, raise: Raise<Error>, _: Ignored)
+      |  fun test() {}
+      |}
+      |"""
+            .trimMargin()
+            .trimMargin()
+
```
hick209, line 237: nit, remove of those calls here

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -7393,6 +7393,63 @@ class FormatterTest {
...
+
+    val expected =
+        """
+      |context(something: Something)
+      |class A {
+      |  context(
+      |  // Test comment.
+      |  logger: Logger,
+      |  raise: Raise<Error>,
+      |  _: Ignored)
```
hick209, line 7433: I'd love to see those indented, but I'm already happy that we at least now have support to this feature

core/pom.xml
```diff
@@ -22,7 +22,7 @@
 
   <properties>
     <dokka.version>0.10.1</dokka.version>
-    <kotlin.version>2.0.21</kotlin.version>
+    <kotlin.version>2.2.0-Beta2</kotlin.version>
```
hick209, line 25: Is this change here necessary? If not and you still want to bump Kotlin let it be its own PR.
Also make sure to touch `online_formatter/build.gradle.kts` as well, like we did in https://github.com/facebook/ktfmt/commit/bcec853c9bd86c529cb285b0376aa35ca37ec3b0.

hick209 (changes requested):

core/pom.xml
ZacSweers: > Is this change here necessary?

Yes, as explained in my PR description:

> Note that in order for `LightVirtualFile` to correctly parse this, it did require updating to the Kotlin 2.2.0 betas. Since this ships as a fat jar I think it would be fine to build against this in main, but up to y'all!

Will split out to a separate PR

ZacSweers: https://github.com/facebook/ktfmt/pull/538

hick209: My bad that I missed that!
In that case you can do the bump here, but up to you now that you already have the 2 PRs open

ZacSweers:
Depends on #538

ZacSweers:
Should be good to go now 🤞

hick209 (approved):

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -1689,12 +1689,19 @@ class KotlinInputAstVisitor(
...
+   * Example `context(logger: Logger, raise: Raise<Error>)`
+   *
+   * Note this also supports the legacy receiver format of `context(Logger, Raise<Error>)` for
+   * backward compatibility.
+   */
   override fun visitContextReceiverList(contextReceiverList: KtContextReceiverList) {
     builder.sync(contextReceiverList)
     builder.token("context")
+    val listToVisit =
+        contextReceiverList.contextParameters().ifEmpty { contextReceiverList.contextReceivers() }
```
hick209, line 1702: This here is being a bit tricky to land internally since `KtContextReceiverList.contextParameters` is not available in the Kotlin version we use internally (2.1.0).

I didn't want to block landing it on the 2.2 upgrade though, but so far I am not sure about what I will do.

I thought about adding code to handle pre-2.2 and 2.1, with the build system doing the conditional here, but I don't know how to (if it is even possible) achieve that with Maven, so first I'd need to move this project to Gradle, which means non-trivial work here.

Let me know in case you have any good ideas on what to do here.

ZacSweers: I don't have good advice unfortunately. The best experience I've had when dealing with internal monorepos is to treat open source projects as prebuilt/external/remote dependencies to be consumed and don't enforce building them internally. Otherwise you get stuck in situations like this where a public project is held back by the company's internal infrastructure.

If the issue is just _merging_ it though, could you merge this to a non-main branch and cut releases from that branch?

hick209: I've pushed an internal PR with support for Gradle, which will allow us to do a modification on top of this PR (which I have ready here so no-op from you) and ship it.

Currently pending internal review.

Thanks a ton for your patience on this one

PR #538, opened 2025-05-05
Update Kotlin to 2.2.0-Beta2

ZacSweers:
Ref: https://github.com/facebook/ktfmt/pull/536#discussion_r2073993355


ZacSweers:
Investigating the failures. So far rootcaused it to a difference in how binary string expressions are represented. Prior to 2.2.0, expressions like `"hello" + "world"` would be tokenized as `KtBinaryExpression`. However, in 2.2.0 they are all passing through as `KtStringTemplateExpression`. The result is the tokenizer has no `+` operators tokenized.  

In formatting they all pass through as `KtBinaryExpression` still, which feels like an inconsistency

ZacSweers:
@hick209 unfortunately I'm not quite sure what to do about this case. Any suggestions? The options seem to be

1. When tokenizing, convert templates back to binary expressions somehow for consistency with the `OpsBuilder` syncs
2. When AST visiting, add the missing op `+` token somehow before visiting the operator
3. If you think this is a bug, report it ASAP to the kotlin issue tracker while it's still in beta
4. ??? something else I'm unaware of but maybe you know?

ZacSweers:
Here's a diff for comparison printing out what is visited in each. I'm going to report this on youtrack as it does seem surprising that it has no reporting of the operator token anymore

image

[2.1.20.txt](https://github.com/user-attachments/files/20049208/2.1.20.txt)
[2.2.0.txt](https://github.com/user-attachments/files/20049209/2.2.0.txt)

ZacSweers:
Filed [KT-77342](https://youtrack.jetbrains.com/issue/KT-77342)

One thing I've considered is implicitly adding a `+` token, but we would also have to then track visitor state to check if the previously visited element was also a string template 🫠

ZacSweers:
@hick209 was able to get it working after a tip in the issue I filed 👍. In short, JB modified the behavior of `KtVisitor.visitBinaryExpression` to _not_ visit all the intermediate elements, but I was able to restore the previous behavior by just overriding this function and having it do the previous implementation. Should be ready to go now 👍

hick209 (approved):
Awesome! Thank you so much for your contributions here and even working with JB on understanding what had changed.
I myself at first thought it was a bug and would have filed an issue in YT just like you did.

I also appreciate your patience on waiting me to get back to you here, this here is a side project of myself so I don't always have time to come and give it the love it deserves

hick209 (changes requested):
Looks like we're still missing those here, sorry I didn't notice and forgot to mention them before
https://github.com/facebook/ktfmt/blob/main/ktfmt_idea_plugin/gradle/libs.versions.toml#L2-L5

I believe this is all of it now, or at least that's my codebase grep told me 😆

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ### Changed
 - All styles managing trailing commas now (https://github.com/facebook/ktfmt/issues/216, https://github.com/facebook/ktfmt/issues/442)
+- Update `kotlin-compiler-embeddable` to `2.2.0-Beta2` for forward compatibility with context parameters.
```
hick209, line 11: While you make the next batch of changes, would you mind including the PR link here as well?

ZacSweers:
> Looks like we're still missing those here, sorry I didn't notice and forgot to mention them before https://github.com/facebook/ktfmt/blob/main/ktfmt_idea_plugin/gradle/libs.versions.toml#L2-L5
> 
> I believe this is all of it now, or at least that's my codebase grep told me 😆

I intentionally left those alone as I assume you wanted to preserve the older version for the intellij runtime, but I can update them if you want and just set it to target kotlin 2.0? let me know

hick209 (approved):

Issue #543, opened 2025-05-19
Intellij .editorconfig has wrong indentation for parameter lists


patricklucas:
I've found that when using the provided .editorconfig for the default configuration in IntelliJ, when typing parameter lists, the parameters are only indented with 2 spaces, whereas when ktfmt runs, that becomes four.

When typing:

```kotlin
class SomeClass {
  fun asdf(
    param1: String,
    param2: String,
    param3: String,
    param4: String,
    param5: String,
    param6: String,
    param7: String,
    param8: String,
    param9: String,
  ) {
    listOf(param1, param2, param3, param4, param5, param6, param7, param8, param9)
        .forEach(::println)
  }
}
```

Formatted by ktfmt to:

```kotlin
class SomeClass {
  fun asdf(
      param1: String,
      param2: String,
      param3: String,
      param4: String,
      param5: String,
      param6: String,
      param7: String,
      param8: String,
      param9: String,
  ) {
    listOf(param1, param2, param3, param4, param5, param6, param7, param8, param9)
        .forEach(::println)
  }
}
```

This seems to be controlled with the EditorConfig option `ij_kotlin_continuation_indent_in_parameter_lists`—setting it to `true` rather than `false` [as it is currently](https://github.com/facebook/ktfmt/blob/v0.54/docs/editorconfig/.editorconfig-default#L40) results in the correct formatting when typing.

I'll open a PR for this, but I'd appreciate it if someone else could confirm this behavior. I don't _think_ the other style configurations are affected by this.

PR #544, opened 2025-05-19
Fix option in .editorconfig-default (#543)

patricklucas:
See context on #543.


hick209 (approved):
I honestly don't fully understand this here, but I trust you did proper testing and validated this helped improve things (at least for you locally)

patricklucas:
> I honestly don't fully understand this here, but I trust you did proper testing and validated this helped improve things (at least for you locally)

Yeah I can't fully explain it either, but unless my ktfmt is behaving strangely (I'm running with Spotless, and use the default style), it definitely indents parameter lists "extra"—4 instead of 2 spaces—and this change certainly seems to get IntellJ to match that.

PR #545, opened 2025-05-21
Fix the GH publish workflow expression syntax

sureshg:


hick209:
Thanks a ton for your contribution here. Unfortunately though I had already an internal PR in review when you published this and it has landed already.

There are still issues with the release though which I'm looking into now

PR #550, opened 2025-06-29
Add --max-width CLI option

ukarlsson:
## Summary
Adds a new `--max-width=<value>` CLI option to allow users to override the default line width (100) for any formatting style.

We are using the Gradle plugin and have it configured at 120 (we have very wide jOOQ queries etc), but now we would also like to use the CLI for faster access since people have started using VSCode with Kotlin LSP and we would prefer to not have that run Gradle. I can of course add the other options available in Gradle plugin if you think that is something that is OK to include.

Many thanks!

## Test plan
- Built and tested the CLI with various max-width values
- Verified error handling for invalid inputs (non-numeric, negative values)
- Confirmed the option appears in help text

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
+++ b/core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
@@ -81,6 +81,7 @@ data class ParsedArgs(
         |  --set-exit-if-changed             Sets exit code to 1 if any input file was not 
         |                                        formatted/touched
         |  --do-not-remove-unused-imports    Leaves all imports in place, even if not used
+        |  --max-width=<value>               Maximum line width (default: 100)
         |  
         |ARGFILE:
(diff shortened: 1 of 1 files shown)


ukarlsson:
@hick209 Hello, any chance we could merge this? 🙏

ukarlsson:
We looked at this a bit more, and we would actually like to have more complete CLI that has support for globs, check/write options and maybe .ktfmtignore. Considering implementation ktfmt-cli package instead and releasing, what do you think?

ukarlsson:
Sorry, I meant that I was considering that we could do that instead if we don't want to extend the CLI that is part of ktfmt with options/features, then we could maintain that separately

ukarlsson:
This is kind of what we would want for our use case

https://github.com/ukarlsson/ktfmt-cli

ukarlsson:
Ok thanks!

PR #555, opened 2025-08-26
Shadow 9.0.2

Goooler:


Goooler:
As there is a bug of diffing large JARs from https://github.com/JakeWharton/diffuse/issues/325, I haven't diffed the artifacts between upgrading or not.

Goooler:
The format validation failure is due to f8fcec67bd09c908bd5e7d9d8023372bec7e2766, not related to my changes.

core/build.gradle.kts
```diff
@@ -109,7 +109,11 @@ tasks {
   }
 
   // Fat jar
-  shadowJar { archiveClassifier.set("with-dependencies") }
+  shadowJar {
+      archiveClassifier = "with-dependencies"
+      duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+      failOnDuplicateEntries = true
```
hick209, line 115: What's the rationale behind adding those 2 lines here?

Goooler: Exclude is the default strategy. But I suggested declaring this property explicitly in configuration at https://github.com/GradleUp/shadow/releases/tag/9.0.1

PR #563, opened 2025-10-09
fix: support when else with guard

tKe:
Currently attempting to format a when expression with an `else if` entry results in:
>  error: expected token: 'else'; generated if instead


hick209:
Thanks a ton for working on this. I'd be happy to merge this, but before could you please rebase and make sure it builds fine and passes the checks?
```
./gradlew build test
```

hick209:
Awesome! Oh, last thing. Could you update `CHANGELOG.md`?

Issue #567, opened 2025-10-27
Align configuration options between CLI, IDE plugin and Gradle


vRallev:
I'm fully aware of the decision to [provide as few configuration options as possible](https://github.com/google/google-java-format/wiki/FAQ#i-just-need-to-configure-it-a-bit-differently-how). I'm also aware of #470 providing a CLI option to change the line length. However, over a year later these decisions have a few bad consequences and I hope the decision will be reconsidered.

The IDE and Gradle plugin both provide the option to change the line length, block indent, continuation indent, trailing comma management and removing unused imports. Through CLI you can only change the unused import option. This inconsistency is a downside as projects started to make use of these options when only formatting through Gradle or the IDE.

In #470 the recommendation was to create wrapper for CLI that provides all these options. Because of this decision, projects like these are born https://github.com/block/kotlin-formatter/ It's basically a ktfmt wrapper only to change the line length. But with a separate CLI binary, there is the valid desire to run all formatting through the same binary, therefore the IDE and Gradle plugin have to be forked. I think this is worse than providing consistent configuration options through all tools.

Another perspective to look at this concern is that code style options are limited to Meta, Google and Kotlin-Lang. Why not allow a company like Block to define their code style for these few options? Allow defining custom options like `--block-style`. 

These configurations could be passed in through CLI in multiple ways. Using an `.editorconfig` as suggested in #45 is another approach.

hick209:
https://github.com/block/kotlin-formatter/ seems like a really good solution from my POV although I do understand where you are coming from here.
At the moment we are still opting not to add support to that to the CLI though, but if things change and this issue is still open, I can come here and let you know.

The existing styles come from historical reasons here.
Meta and kotlin lang: Because ktfmt was developed for use at Meta, where we use 2-space indent which diverges from kotlin-lang recommendations. That created the "first config fork".
Google decided to use ktfmt and joined the effort, helping implement various features. They wanted some special treatment for their codebase and we didn't want to add support for that. They agreed to develop it and provide support for it – which they have been doing – and that's how the Google style became a thing.

If Blocks needs something similar, then maybe it could be arranged. Feel free to hit me up on the slack channel if you want to have further discussions around this.

I do like something like #45 and we could definitely take a PR around that if anyone is willing to contribute.

The main idea here is to haver the formatting easily reproducible.
[...]

PR #570, opened 2025-11-12
feat: --enable-editorconfig flag and .editorconfig support

### LINKED ISSUE #45: Consume .editorconfig for configuration
https://editorconfig.org/ is a simple standard that both KTLint and Intellij can consume from. This would be a nice addition to ktfmt to allow for more configuration when needed without custom APIs/params in the tool.

tKe:
closes #45


tKe:
@hick209 is this inline with you were alluding to in your [comment](https://github.com/facebook/ktfmt/issues/567#issuecomment-3514025111) in #567?

hick209:
There are some minor things that I'll fix myself here before landing. Awesome work @tKe, thanks for contributing to making ktfmt better for everyone

PR #571, opened 2025-11-19
Explicit Kotlin import layout

vRallev:
Specify an explicit import layout instead of relying on the Kotlin Official default. The Kotlin Official style uses `ij_kotlin_imports_layout = *,java.**,javax.**,kotlin.**,^`. The new explicit import layout changes it to how ktfmt formats code.

`.editorconfig-kotlinlang` has already applied the same setting.


hick209:
Thanks for contributing! If you don't mind (or I don't do it first) would you mind updating `CHANGELOG.md`?

cortinico (approved):

PR #574, opened 2025-12-03
Issue 547 cascade nested lambda breaks

### LINKED ISSUE #547: Reconsider lambda single-lining for Jetpack Compose
I know this is going to be controversial & I'm a fan of single config, but in 2025 UI's are made with Kotlin as well (Jetpack Compose), and trailing lambda is very common,

however in UIs nesting means..nesting. It has a semantic meaning (parent-child relationship). It's not just a stylistic choice.

```
App { SelectableCard { Button { Text("") }}}
```

Is obviously not conveying that & makes it harder to reason about, versus

```
App { 
    SelectableCard { 
        Button { 
            Text("")
        } 
    } 
}
```
which is natural way of reasoning about UIs

loganj:


hick209 (changes requested):
As mentioned above, let's have the option to preserve lambda/block line breaks instead as this is more general purpose.

PR #582, opened 2025-12-23
fix: editorconfig not found for relative paths

tKe:
In the common case of a `.editorconfig` in the root of the repo or
project, when running `ktfmt` on a path like `src/main/kotlin/MyFile.kt`
it will stop looking in `src` as `src` has no "parent" and not find any
`.editorconfig`. Using a path like `./src/main/kotlin/MyFile.kt` will
stop at `.` and find `./.editorconfig`.

By using the absolute path of the file, it'll walk all ancestors until
it finds a `.editorconfig` with `root = true` as expected.

I've also added `ktfmt_trailing_comma_management_strategy` to the example editorconfigs and fixed an issue with the gradle build supporting [configuration-cache](https://docs.gradle.org/current/userguide/configuration_cache.html). I'm happy to move either/both of these to separate PRs if required.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/cli/EditorConfigResolver.kt
+++ b/core/src/main/java/com/facebook/ktfmt/cli/EditorConfigResolver.kt
@@ -20,6 +20,7 @@ import com.facebook.ktfmt.format.FormattingOptions
 import com.facebook.ktfmt.format.TrailingCommaManagementStrategy
 import java.io.File
 import java.util.concurrent.ConcurrentHashMap
+import kotlin.io.path.absolute
 import org.ec4j.core.EditorConfigLoader
 import org.ec4j.core.PropertyTypeRegistry
 import org.ec4j.core.Resource
@@ -28,7 +29,6 @@ import org.ec4j.core.ResourcePropertiesService
 import org.ec4j.core.model.EditorConfig
 import org.ec4j.core.model.PropertyType
 import org.ec4j.core.model.Version.CURRENT
-import kotlin.io.path.absolute
 
 object EditorConfigResolver {
 


core/build.gradle.kts
```diff
@@ -61,14 +61,14 @@ tasks {
...
-            commandLine = listOf(genVersionFileScript.toString(), versionPropertiesFile.toString())
-          }
+    // provider to run the shell script genVersionFileScript with versionPropertiesFile as argument
+    val scriptProcess =
+        providers.exec {
+          workingDir = rootProject.rootDir
+          commandLine = listOf(genVersionFileScript.toString(), versionPropertiesFile.toString())
+        }
 
+    doLast {
```
hick209: Could you clarify the rationale for this change here?

tKe: This avoids the following error when running gradle with configuration-cache enabled:

```
1 problem was found storing the configuration cache.
- Task `:ktfmt:generateKtfmtFile` of type `org.gradle.api.DefaultTask`: cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.
  See https://docs.gradle.org/8.14/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
```

The `providers` property is accessing the `Project` during task execution within `doLast` which is [no longer permitted](https://docs.gradle.org/current/userguide/configuration_cache_requirements.html#config_cache:requirements:disallowed_types) when configuration cache is enabled. Given the provider is lazily evaluated, it's safe to move it outside the `doLast` to be lazily evaluated during the task execution phase.

It's not required by any means, especially since the build still uses 8.14, but from 9.0 onwards enabling the configuration cache is the [preferred mode of execution](https://docs.gradle.org/9.0.0/release-notes.html#config-cache) (although it remains off by default).

PR #584, opened 2026-01-12
Make `ktfmt` up to 100x faster via `native-image`

### LINKED ISSUE #441: Feature request: GraalVM artifacts
GJF just added this support to their releases: https://github.com/google/google-java-format/releases/tag/v1.20.0

sgammon:
## Native ktfmt

Introduces support for a native binary version of `ktfmt`, via [GraalVM `native-image`](https://www.graalvm.org/latest/reference-manual/native-image/); no code changes have taken place, just build changes, in order to make kotlinc's infrastructure (which ktfmt depends on) usable under SVM, Substrate Virtual Machine.

After some optimization and cleanup, I have some initial numbers which show promising gains:

image

The native version of `ktfmt` is identical in every way to the regular JVM version, except it is built AOT, and, since it is already machine code, it does not need a JVM installed at runtime.

We have been using this version of ktfmt internally for some time now -- a few weeks -- and it is stable for us even on codebases with hundreds or thousands of files. Once you hit many thousands of files, the performance gap between JVM and SVM begins to close (this is tunable). That makes sense, considering JVM's warmed-up JIT state is still known to beat SVM at peak performance.

However, for smaller executions of ktfmt -- particularly ones which only format changed code -- the difference can be quite persuasive toward native, because of essentially instant startup time.

Note: To actually ship this on ktfmt, we are going to need build infrastructure for each target OS/architecture pair. GraalVM's support matrix currently includes:
- Linux amd64 / arm64
- macOS amd64 / arm64
- Windows amd64

I am able to build `ktfmt` for each of the above targets, and have CI (personally) to help, but will need to transition that stuff to GHA and Meta's control. In conversations with Nivaldo (@hick209), he indicated CI should be no problem.

Fixes and closes #441 cc / @ZacSweers 

## Trying it out

Here are pre-built binaries for Linux amd64 and macOS arm64, if you'd like to try it out. Each package comes with `ktfmt` and `ktfmt-build-report.html`, which is a standalone file you can open in your browser. It shows the full contents of the native image and various stats about its contents.

[Download for macOS arm64](https://static.elideusercontent.com/ktfmt.macos-arm64.txz)
[Download for Linux amd64](https://static.elideusercontent.com/ktfmt.linux-amd64.txz)

The final compressed packages end up at about `13MB` for macOS and `16MB` for Linux.

image

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -28,6 +28,11 @@ jobs:
       with:
         java-version: ${{ matrix.java }}
         distribution: zulu
+    # set-java-home is intentionally 'false': the main build runs on the Zulu JDK from the matrix
+    # (Java 17/21), so we must not repoint JAVA_HOME at GraalVM 25. setup-graalvm still exports
+    # GRAALVM_HOME, which the GraalVM Gradle plugin uses to locate the native-image toolchain for the
+    # :ktfmt:nativeCompile step. This keeps the JVM build on the matrix JDK and the native build on
+    # GraalVM.
     - name: Set up GraalVM
       uses: graalvm/setup-graalvm@v1
       with:
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -4,6 +4,8 @@ on:
   push:
     paths-ignore:
       - '**.md'
+    branches:
+      - main
   pull_request:
     paths-ignore:
       - '**.md'
@@ -25,5 +27,17 @@ jobs:
       with:
         java-version: ${{ matrix.java }}
         distribution: zulu
+    - name: Set up GraalVM
+      uses: graalvm/setup-graalvm@v1
+      with:
+        java-version: '25.0.1'
+        distribution: 'graalvm'
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        set-java-home: 'false'
+        native-image-job-reports: 'true'
     - name: Build ktfmt
       run: ./gradlew build --no-daemon
+    - name: Build Native Image
+      run: ./gradlew :ktfmt:nativeCompile --stacktrace --no-daemon
+    - name: Smoke test Native Image
+      run: ./native_smoke_test.sh
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -6,6 +6,7 @@ on:
       - '**.md'
     branches:
       - main
+      - feat/native-image
   pull_request:
     paths-ignore:
       - '**.md'
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -28,6 +28,7 @@ plugins {
   id("maven-publish")
   id("signing")
   id("ktfmt.ktfmt-file-generator")
+  id("ktfmt.native-image")
 }
 
 repositories {
--- a/core/src/main/java/com/facebook/ktfmt/format/Parser.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/Parser.kt
@@ -43,9 +43,7 @@ object Parser {
    * from [KotlinCoreEnvironment.createForProduction]:
    * https://github.com/JetBrains/kotlin/blob/master/compiler/cli/src/org/jetbrains/kotlin/cli/jvm/compiler/KotlinCoreEnvironment.kt#L544
    */
-  val env: KotlinCoreEnvironment
-
-  init {
+  val env: KotlinCoreEnvironment by lazy {
     // To hide annoying warning on Windows
     System.setProperty("idea.use.native.fs.for.win", "false")
     val disposable = Disposer.newDisposable()
@@ -54,13 +52,12 @@ object Parser {
         CommonConfigurationKeys.MESSAGE_COLLECTOR_KEY,
         PrintingMessageCollector(System.err, PLAIN_RELATIVE_PATHS, false),
     )
-    env =
-        @Suppress("OPT_IN_USAGE_ERROR") // KotlinCoreEnvironment.createForProduction
-        KotlinCoreEnvironment.createForProduction(
-            disposable,
-            configuration,
-            EnvironmentConfigFiles.JVM_CONFIG_FILES,
-        )
+    @Suppress("OPT_IN_USAGE_ERROR") // KotlinCoreEnvironment.createForProduction
+    KotlinCoreEnvironment.createForProduction(
+        disposable,
+        configuration,
+        EnvironmentConfigFiles.JVM_CONFIG_FILES,
+    )
   }
 
   fun parse(code: String): KtFile {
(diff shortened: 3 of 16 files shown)


core/src/main/native-image/profiles/README.md
```diff
@@ -0,0 +1,5 @@
+
+### PGO Profiles
+
+Run `pgo_train.sh` to generate `default.iprof` in this folder. Then, it will be used when `-Pktfmt.native.pgo=true` is passed to `./gradlew :ktfmt:nativeCompile`.
```
sgammon: Instructions to build and use PGO (Profile Guided Optimization) profiles are included, as well as profiles themselves. PGO significantly improves startup and peak performance

core/src/main/java/com/facebook/ktfmt/format/Parser.kt
```diff
@@ -54,12 +52,11 @@ object Parser {
...
-        KotlinCoreEnvironment.createForProduction(
-            disposable,
-            configuration,
-            EnvironmentConfigFiles.JVM_CONFIG_FILES,
-        )
+    KotlinCoreEnvironment.createForProduction(
+        disposable,
+        configuration,
+        EnvironmentConfigFiles.JVM_CONFIG_FILES,
+    )
```
sgammon: the _only_ code change: `val env` was made `lazy`, so that it is not initialized at build time; otherwise, the native image size would grow considerably for no benefit. has no impact to JVM mode.

sgammon (commented):
self-review

core/src/main/java/com/facebook/ktfmt/format/Parser.kt
hick209: SGTM. I'm now curious why would lazy here impact the image size? 🤔

core/build.gradle.kts
```diff
@@ -137,6 +186,181 @@ ktfmt {
   )
 }
 
+object DefaultArchitectureTarget {
+  val amd64 = "x86-64-v4"
+  val arm64 = "armv8.4-a+crypto+sve"
+}
```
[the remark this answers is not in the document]
hick209: That part here I"m just going to trust you as it's beyond my knowledge domain.

What about archs like x86? I then see that you funnel that to amd64 below. Wouldn't it be better to explicit about that here? Even if just a simple `val x86_64 = amd64`

core/build.gradle.kts
```diff
@@ -137,6 +186,181 @@ ktfmt {
...
+          "amd64",
+          "x86_64" -> DefaultArchitectureTarget.amd64
+          "aarch64",
+          "arm64" -> DefaultArchitectureTarget.arm64
+          else -> error("Unrecognized host architecture: '$hostArch'")
+        }
+
+// Pass `-Pktfmt.native.gc=xx` to select a garbage collector; options include `serial`, `G1`, and
+// `epsilon`.
+val nativeGc = findProperty("ktfmt.native.gc") ?: "G1"
```
hick209: That's not my domain: Isn't "Serial GC" the community one? Could there be licensing issues with G1?

core/build.gradle.kts
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
```
hick209: Overall I see a ton of changes here. Would it be possible to put them (at least a good chunk) into their own file so they do not "pollute as much" the build files? At the very least try to section them off somehow.

WDYT?

hick209 (commented):
Initial batch. There shall be more tomorrow 😃

Here are my initial comments. No need to act on anything just yet if you do not desire.

core/src/main/native-image/java/com/facebook/ktfmt/nativeImage/KotlinCoreEnvironmentCompanion.java
```diff
@@ -0,0 +1,34 @@
...
+@SuppressWarnings({"unused"})
+@TargetClass(KotlinCoreEnvironment.Companion.class)
+final class KotlinCoreEnvironmentCompanion {
+    @Substitute
+    @TargetElement(name = "registerApplicationExtensionPointsAndExtensionsFrom")
+    private void stubbedRegisterApplicationExtensionPointsAndExtensionsFrom(
+            CompilerConfiguration configuration, String configFilePath) {
+        // Nothing at this time.
+    }
+}
```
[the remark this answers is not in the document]
sgammon: Often, yes, because one must target exact types with no room for error. I can try this in Kotlin to see if it works; a type misalignment should be picked up by Native Image. But it's often best to express in Java for this reason.

This is the only code that would ever need to be held in Java. It won't need to evolve or be updated, really, until the Kotlin compiler team finishes their native work upstream, at which point it can simply be deleted.

hick209 (commented):
It's honestly a lot of things I don't really understand here.
From my limited understanding this seems like a good improvement and I'm tempted to just merge it in and give it a go.

Right now I see there some CI jobs (here from GH) failed, could you take a look to see if any of that makes sense before I pull this in for internal review?

eduardb:
Was this PR abandoned? 😢

sgammon:
@eduardb not abandoned, no, just trying to figure out windows CI; meta must facilitate windows and mac CI in order to support this on all operating systems. we have offered to provide those CI services at our own cost.

i will be updating this to match `main`

sgammon:
if, in the meantime, any community members would like `ktfmt` in native mode, please use the downloads in this PR, or use [Elide](https://elide.dev), which has `ktfmt` built in at `elide ktfmt -- ...`. It's free. this is the only way we know how to provide this download to users before this PR is merged.

hick209:
Is the build break there expected?

Not the CodeQL one, that one seem broken for all commits, so don't worry

sgammon:
@hick209 now the only broken step is codeql

sgammon:
@hick209 OK, I have split it up, rebased, and cleaned it up a bit:

1) **`fix: lazily initialize Kotlin parser environment`**
Isolates the only actual code change; `Parser.env` is now initialized lazily, which helps `native-image` avoid performing Kotlin's compiler setup too early. That should happen at runtime, rather than build time. There is effectively zero behavior change on JVM.

2) **`feat: build ktfmt native image`**
What changed:
- Moved native image build logic to a convention plugin, so it is isolated and doesn't pollute the main build logic
- Moved native image libraries to a dedicated version catalog, at `gradle/native-image.versions.toml`
- Split PGO out into another branch, so it can be added later (this removes a lot of stuff)
- Marks reachability metadata as generated in `.gitattributes` so it doesn't pollute review

3) **`chore: run CI on native-image branch`**
This is just a temporary commit to get CI to run [on my fork](https://github.com/sgammon/ktfmt/actions/runs/27716420350). I intend to test there a bit as I go. This commit should be dropped before merge

.github/workflows/build_and_test.yml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     paths-ignore:
       - '**.md'
+    branches:
+      - main
+      - feat/native-image
```
hick209, line 9: We only plan to keep the main branch, so I guess we can drop this here, right?

hick209:
Nice! It looks much better / cleaner now! Thank you

.github/workflows/build_and_test.yml
sgammon: Yes, this is a temporary commit only, so that testing can happen on my fork; I can make sure the build is green before pinging you for a build approval

hick209 (approved):

sgammon:
thank you for your guidance on this PR @hick209, and for your support on your amendment PR, @ZacSweers

PR #613, opened 2026-04-26
Remove forced breaking of contextual function types

### LINKED ISSUE #612: Contextual function types are broken apart as if they were declarations
I suspect this is because the context list forcefully goes on a separate line, even if it's actually part of a type.
I understand this could be a matter of taste, but just look at these examples, they're awful!
(I had to run this locally since it errors on the website)
```kotlin
// input
fun foo(block: context(Foo) () -> Unit) {}
// output
fun foo(
  block:
      context(Unit)
      () -> Unit
) {}
```
```kotlin
// input
context(_: Fresh)
suspend fun bindHere2(block: suspend context(Bind) () -> Stmt): Stmt = TODO()
// output
context(_: Fresh)
suspend fun bindHere2(
  block:
    suspend context(Bind)
    () -> Stmt
): Stmt = TODO()
```
The expectation on my end is that both inputs are properly formatted (note that if I were using extension lambdas, that is the case)
It's making me question whether I should use ktfmt in a codebase with context parameters at all.

kyay10:
Fixes #612 

A rather simple change, ultimately. It seems that the previous behaviour was "intentional", but I think it's rather incorrect. Given that this is an experimental feature, I think changing it now is better than waiting for more complaints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 ### Added
 - Support `ij_kotlin_indent_size` in editorconfig. (https://github.com/facebook/ktfmt/pull/604)
 - Support for lists within quoted blocks in KDoc comments (https://github.com/facebook/ktfmt/commit/68fa1585b759ad4b12ca4802bccd297f6a33b0f3)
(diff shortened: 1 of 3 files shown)


hick209 (approved):

hick209 (changes requested):
Please do not forget to update the `CHANGELOG.md`

hick209:
Thanks for the quick response

cortinico (approved):

PR #614, opened 2026-04-28
Default to preserving lambda line breaks

### LINKED ISSUE #547: Reconsider lambda single-lining for Jetpack Compose

ZacSweers:
This is a continuation of #574 with the updated option control to orient around the requested `--preserve-lambda-breaks` simple option. With this, lambda block line breaks are simply just respected rather than have ktfmt impose anything.

Aside from finishing up the previous PR, this also adds a setting to the IJ plugin.

I cherrypicked the original commits to preserve authorship history (CC @loganj)

Resolves #547

**Edit**

After chatting on slack I updated the PR with a couple things:
- Added a builder API for CLI tools to use in the future to avoid ABI breakages
- Flipped the semantics to enable this by default, but can revert if you want to keep the old setup. I do think enabled by default is good


ZacSweers:
The CI failure looks like a bootstrap compatibility issue - can't add a new param without breaking ktfmt-gradle since it's statically compiled against the only one? I tried adding `@JvmOverloads`  to the constructor but no dice on that either, it's a chicken-and-egg problem 😅 🥚

core/src/main/java/com/facebook/ktfmt/format/FormattingOptions.kt
hick209: For the CI failure, what if we added a new constructor here?
```
data class FormattingOptions(
  ...
) {
  ...
  @Deprecated("Here just for retrocompatibility reasons. Will be removed on 1.0.0")
  constructor(
    maxWidth: Int,
    blockIndent: Int,
    continuationIndent: Int,
    trailingCommaManagementStrategy: TrailingCommaManagementStrategy,
    removeUnusedImports: Boolean,
    debuggingPrintOpsAfterFormatting: Boolean,
  ) : this(
    maxWidth = maxWidth,
    blockIndent = blockIndent,
    continuationIndent = continuationIndent,
    trailingCommaManagementStrategy = trailingCommaManagementStrategy,
    removeUnusedImports = removeUnusedImports,
    preserveLambdaBreaks = false,
    debuggingPrintOpsAfterFormatting = debuggingPrintOpsAfterFormatting,
  )
  ...
}
```

hick209 (commented):
Love the work you did here! Thank you so much and sorry it took so long to get to this.

I am not fully confident on exposing this via CLI flag yet, everything else LGTM.
Regardless I'll pull it internally for review.

hick209 (changes requested):
Oh, please do not forget to update the CHANGELOG.md

core/src/main/java/com/facebook/ktfmt/format/FormattingOptions.kt
ZacSweers: I think that could do it, let me try. My hope was that the `@JvmOverloads` option would cover it but for some reason that didn't do the trick before

ZacSweers:
Updated the changelog and added another constructor overload. For the latter - it may be valuable before 1.0.0 to expose a Builder API for this to avoid the churn for tools programmatically calling this

ZacSweers:
Doesn't look like we can get away with an overload

```
Execution failed for task ':ktfmt:ktfmtFormatMain'.
> There were multiple failures while executing work items
   > A failure occurred while executing com.ncorti.ktfmt.gradle.tasks.worker.KtfmtWorkAction
      > 'void com.facebook.ktfmt.format.FormattingOptions.<init>(int, int, int, com.facebook.ktfmt.format.TrailingCommaManagementStrategy, boolean, boolean)'
   > A failure occurred while executing com.ncorti.ktfmt.gradle.tasks.worker.KtfmtWorkAction
      > 'void com.facebook.ktfmt.format.FormattingOptions.<init>(int, int, int, com.facebook.ktfmt.format.TrailingCommaManagementStrategy, boolean, boolean)'
   > A failure occurred while executing com.ncorti.ktfmt.gradle.tasks.worker.KtfmtWorkAction
      > 'void com.facebook.ktfmt.format.FormattingOptions.<init>(int, int, int, com.facebook.ktfmt.format.TrailingCommaManagementStrategy, boolean, boolean)'
   > A failure occurred while executing com.ncorti.ktfmt.gradle.tasks.worker.KtfmtWorkAction
      > 'void com.facebook.ktfmt.format.FormattingOptions.<init>(int, int, int, com.facebook.ktfmt.format.TrailingCommaManagementStrategy, boolean, boolean)'
   > A failure occurred while executing com.ncorti.ktfmt.gradle.tasks.worker.KtfmtWorkAction
      > 'void com.facebook.ktfmt.format.FormattingOptions.<init>(int, int, int, com.facebook.ktfmt.format.TrailingCommaManagementStrategy, boolean, boolean)'
   > A failure occurred while executing com.ncorti.ktfmt.gradle.tasks.worker.KtfmtWorkAction
      > 'void com.facebook.ktfmt.format.FormattingOptions.<init>(int, int, int, com.facebook.ktfmt.format.TrailingCommaManagementStrategy, boolean, boolean)'
```

hick209:
Debugged offline via Slack

hick209 (approved):

core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
```diff
@@ -130,6 +134,7 @@ data class ParsedArgs(
           arg == "--do-not-remove-unused-imports" -> removeUnusedImports = false
           arg == "--enable-editorconfig" -> editorConfig = true
           arg == "--quiet" -> quiet = true
+          arg == "--do-not-preserve-lambda-breaks" -> preserveLambdaBreaks = false
```
hick209: My last request here would be to remove this CLI option. Everything else can remain the same

hick209 (changes requested):

core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
ZacSweers: Done! [3c5f5fd](https://github.com/facebook/ktfmt/pull/614/commits/3c5f5fd42f53812a079e7cf861c0754367758ea8)

hick209 (approved):

PR #617, opened 2026-05-25
Finish GraalVM `native-image` support

### LINKED ISSUE #441: Feature request: GraalVM artifacts

ZacSweers:
This finishes up #584. The original commits are preserved. Closes #441.

### Changes since #584

- Rebased onto current `main`
- Regenerated the reachability metadata for the current Kotlin compiler version and fixed build-time vs. runtime class initialization so the image builds and runs (was crashing on the newer kotlinc)
- Default to the Serial GC instead of G1. G1 is unsupported by `native-image` on macOS and is Oracle  GraalVM only; Serial fits ktfmt's short-lived runs. Opt into G1 with `-Pktfmt.native.gc=G1`
- Use portable `-march` defaults (`x86-64-v3`, `armv8-a`). The old arm64 value was not a valid
  `native-image` target, and `x86-64-v4` would crash on older/CI CPUs. Matches what GJF does.
- Refactored the native-image build logic into a `build-logic` convention plugin
- Rewrote the Java SVM substitution in Kotlin.
- Dropped the unnecessary `application` plugin
- A missing PGO profile now degrades to a non-PGO build instead of failing
- Build the native image in its own CI job across linux/macOS/Windows with a smoke test that actually formats code so stale reflection metadata is caught early
- Added a release job that builds and uploads a native binary per platform, modeled on GJF's release workflow
- Added a README section on the native binary
- `native_smoke_test.sh` and `regenerate_native_metadata.sh` helper scripts and a `.gitattributes` that marks the generated metadata so reviews collapse it
- The cross-platform CI matrices and the release upload still need a real run to shake out, Windows-native especially.
- PGO profiles in CI and release checksums are possible follow-ups


ZacSweers:
Sorry for the churn, doing a bit of leg work to make the repo build able in windows for this.

core/build.gradle.kts
```diff
@@ -52,28 +72,44 @@ val generateSources by tasks.registering {
...
+           * limitations under the License.
+           */
+
+          package com.facebook.ktfmt.util
+
+          object Ktfmt {
+            const val version = "${version.get()}"
+          }
+          """
+                  .trimIndent() + "\n"
```
hick209: Indentation here could be improved to align with the `"""`

gen_version_file.sh
```diff
@@ -1,50 +0,0 @@
-#!/bin/sh
```
hick209, line 1: We still need this file as it's used by our internal BUCK build.
I guess I could move it around so it's publicly hidden though...

ZacSweers: Gotcha, yeah feel free to add back whatever you all need internally!

core/build.gradle.kts
ZacSweers:
```suggestion
          """
          .trimIndent() + "\n"
```

hick209:
Favoring original #584.
Feel free to reopen this if you want to land any of the other changes in this diff (after rebasing it)

PR #620, opened 2026-05-30
Avoid redundant allocations

### LINKED ISSUE #619: Low-hanging fruits in ktfmt performance

qwwdfsad:
* Optimizes work with `PsiElement`: no direct `text` access where appropriate as it's relatively expensive
* Some early returns to avoid String copying when it is not necessary
* Around 6-7% performance improvements on my machine on sequential `kotlinx.coroutines` runs (5 -> 4.65 sec in sequantial mode)

Partially addresses #619

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ### Changed
 
+* Reduced overall number of allocations to improve formatting performance (~6-7%) (https://github.com/facebook/ktfmt/pull/620)
 
 ## [0.63]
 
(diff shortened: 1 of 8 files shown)


qwwdfsad:
@AbdullinAM PTAL

hick209:
Could you update the [`CHANGELOG.md`](https://github.com/facebook/ktfmt/blob/main/CHANGELOG.md)?

hick209 (approved):

qwwdfsad:
Sure, done

AbdullinAM (commented):
Nice work! Note for the future: we should add some benchmarks in this repo, to track these kind of performance gains

cortinico (approved):

PR #622, opened 2026-06-03
Reuse Parser.parse results

### LINKED ISSUE #552: Four calls to `Parser.parse` cause performance issues
Hey! We've been using ktfmt for our generated SDKs at [Stainless](https://stainless.com). We've found the formatter can be slow sometimes due to the size of the SDKs we generate (the larger the API, the larger the generated SDK).

I decided to do some profiling and I found `Parser.parse` takes a significant amount of time ([ktfmt-flamegraph.html.zip](https://github.com/user-attachments/files/21414631/ktfmt-flamegraph.html.zip) attached for running [`anthropic-sdk-java`](https://github.com/anthropics/anthropic-sdk-java)) because it's called [four times](https://github.com/search?q=repo%3Afacebook%2Fktfmt+%22Parser.parse%28%22+-path%3Atest&type=code) per file.

Do you think it's possible to ensure each file gets parsed exactly once? I recognize it's not a trivial problem because subsequent transformations would somehow need to take into account edits from previous transformations, but seems like a worthwhile investment.

### LINKED ISSUE #619: Low-hanging fruits in ktfmt performance

qwwdfsad:
Cuts down formatting time on coroutines from 4.7 to ~2.9-3 seconds. 
All explanations are in commit messages (note: they should be reviewed separately, the first one addresses a non-functional but necessary change)

Partially addresses #619
Fixes #552

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/format/FormatterContext.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/FormatterContext.kt
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.facebook.ktfmt.format
+
+import org.jetbrains.kotlin.psi.KtFile
+
+internal class FormatterContext(@JvmField val code: String) {
+
+  private var _ktFile: KtFile? = null
+  private val ktFile: KtFile
+    get() = _ktFile ?: Parser.parse(code).also { _ktFile = it }
+
+  inline fun transform(block: (KtFile) -> String): FormatterContext {
+    val newCode = block(ktFile)
+    return if (newCode == code) this else FormatterContext(newCode)
+  }
+}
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/facebook/ktfmt/format/FormatterContext.kt
```diff
@@ -0,0 +1,31 @@
...
+
+package com.facebook.ktfmt.format
+
+import org.jetbrains.kotlin.psi.KtFile
+
+internal class FormatterContext(@JvmField val code: String) {
+
+  private var _ktFile: KtFile? = null
+  private val ktFile: KtFile
+    get() = _ktFile ?: Parser.parse(code).also { _ktFile = it }
```
hick209, line 25: Couldn't we just use lazy initialization here?

core/src/main/java/com/facebook/ktfmt/format/RedundantElementManager.kt
```diff
@@ -33,8 +34,11 @@ import org.jetbrains.kotlin.psi.psiUtil.startOffset
  */
 object RedundantElementManager {
   /** Remove extra semicolons and unused imports, if enabled in the [options] */
-  fun dropRedundantElements(code: String, options: FormattingOptions): String {
-    val file = Parser.parse(code)
+  fun dropRedundantElements(code: String, options: FormattingOptions): String =
+      dropRedundantElements(Parser.parse(code), options)
```
hick209: In case we no longer need those functions (if unused, which from my quick grep here seems to be the case) feel free to remove them.
Same applies to the other functions where you made similar changes.

hick209 (approved):

core/src/main/java/com/facebook/ktfmt/format/FormatterContext.kt
qwwdfsad: Not sure I follow here -- this `var` can change multiple times (and `parse` will be invoked multiple times) if an intermediate formattingstep will change the code, so `ktFile by lazy { ... }` isn't enough. Or did you mean something else here?

core/src/main/java/com/facebook/ktfmt/format/RedundantElementManager.kt
qwwdfsad: Right now, these are a part of the public ABI (that is being kept track of), thus I haven't removed them (and new overloads are `internal` to avoid adding new things) as wasn't sure if it's acceptable

I can remove all of the previous overloads if it's okay, and it's not actually part of the public ktfmt API/contract.

qwwdfsad: Done, as a separate commit

core/src/main/java/com/facebook/ktfmt/format/FormatterContext.kt
hick209: It certainly could change multiple times, but from the current code it does not and it's private.
Unless you have future plans (or I'm missing something here), using lazy here would actually be the same
```
private val ktFile: KtFile by lazy(...) { Parser.parse(code) }
```

cortinico (approved):

PR #623, opened 2026-06-03
Refactor build scripts

### LINKED ISSUE #619: Low-hanging fruits in ktfmt performance

### LINKED ISSUE #621: Refactor build scripts
Inspired by #619 and in preparation for #616, I want to rework some things with the build of the ktfmt. Right now I have four things this I want to change:

* [x] Remove the cyclic task dependencies in `:ktfmt` module; make it so the whole project can be build with just `./gradlew build` and remove the `smoke_tests.sh` script
* [x] Replace the uses of OS-dependent path APIs in build files
* [x] Replace the OS-dependent `gen_version_file.sh` script with a custom Gradle task/plugin
* [x] Migrate from deprecated Dokka V1 Gradle plugin to V2

Do these things sound reasonable to you? If so, I can start working on them

AbdullinAM:
Several fixes related to the build configuration of the project:

* Make it so the whole project can be build via `./gradlew build`
* Add `GenerateKtfmtFilePlugin` plugin that replaces `gen_version_file.sh`
* Update to the latest Dokka, fix Dokka/KDoc related warnings

Each commit is relatively independent, but mostly changes the same files. So I didn't split this into separate PRs for now, but can do so if necessary.

Fixes #621


buildSrc/src/main/kotlin/com/facebook/ktfmt/GenerateKtfmtFileTask.kt
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+
+package com.facebook.ktfmt.util
+
+object Ktfmt {
+${properties.toList().joinToString("\n") { "    const val ${it.first} = \"${it.second}\"" }}
+}
+            """
+            .trimIndent()
```
hick209: What's up with the indent here? Is it not covered by ktfmt? If so, could you update it so that changes?

hick209 (approved):
Mostly good, just one nit

buildSrc/src/main/kotlin/com/facebook/ktfmt/GenerateKtfmtFileTask.kt
AbdullinAM: Yes, for some reason ktfmt accept this formatting. I'll update it manually

hick209: Oh, I know why!
It uses interpolation, so we cannot adjust the indent without risking messing up with the result.
You could extract the interpolation to add it after the `trimIndent()` or something like that if you want ktfmt to take care of the formatting.

AbdullinAM:
@hick209 Hey! Can you comment on the status of this PR? As far as I understand, the internal CI checks have faield (probably because I changed the build setup), so where do we go from here?

hick209:
Hey @AbdullinAM, sorry I went on a week-long vacation. I'm back today and will try to merge this now!

PR #624, opened 2026-06-04
#461 — Add commas to single-item lists

### LINKED ISSUE #461: Trailing comma gets removed if there is only one parameter and it is formatted to be on its own line
Using the ktfmt IntelliJ plugin 1.1.0.47 in Android Studio, setting "Code style" to "Google (internal)"
image
image

1. The following code with a single parameter will get the trailing comma removed.
```
import android.content.SharedPreferences

private class AllowlistedDiskReadsSharedPreferences(
  private val wrappedPreferences: SharedPreferences,
) : SharedPreferences by wrappedPreferences
```

2. This code with two parameters will not.
```
import android.content.SharedPreferences

private class AllowlistedDiskReadsSharedPreferences(
  private val wrappedPreferences: SharedPreferences,
  private val wrappedPreferences2: SharedPreferences,
) : SharedPreferences by wrappedPreferences 
```

Since one of the reasons of using trailing comma is to avoid having to modify an existing line when adding another parameter (https://kotlinlang.org/docs/coding-conventions.html#trailing-commas), perhaps even when there is only one parameter but it has to be on its own line, it should also have a trailing comma?

Thanks!

lwasyl:
Fixes #461

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ### Changed
 
+- Trailing comma is now added when a single parameter is formatted onto its own line (https://github.com/facebook/ktfmt/issues/461) 
+
 ### Fixed
 
 * Fix non-idempotent formatting when a managed trailing comma pushes a line over MAX_WIDTH (e.g. a long qualified expression as the last argument of a call). The comma is now accounted for by re-running the layout, so the line is broken correctly on the first pass. (https://github.com/facebook/ktfmt/pull/636)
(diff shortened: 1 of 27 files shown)


hick209:
@nreid260 thoughts?

lwasyl:
@hick209 just for context, since I haven't posted in the issue — I felt there's general consensus that these trailing commas would be useful, so figured I'd open the PR in case it was just that nobody got around to implementing it

hick209:
Rebase, update `CHANGELOG.md` and we might take this in

lwasyl:
@hick209 done!

lwasyl:
@hick209 looks like some of the checks failed there, any idea why? Happy to add some tests and iterate on this PR if the fix is not sufficient

hick209:
Just some internal tests. I'll work through them.
Don't worry about the rebase either, I've fixed that here internally

cortinico (approved):

PR #626, opened 2026-06-08
Fix trailing lambda with chained call indentation (#589)

### LINKED ISSUE #589: Strange formatting for lambda with chained function call
(kotlinlang-style)
```kotlin
fun quux() =
    runnnnn {
            foo()
            bar()
        }
        .baz()
```
(Java-like)
```kotlin
fun quux() =
    runnnnn {
          foo()
          bar()
        }
        .baz()
```
```kotlin
fun quux() {
    runnnnn {
            foo()
            bar()
        }
        .baz()
}
```
```kotlin
fun quux() {
    runnnnn {
        foo()
        bar()
    }.baz()
}
```
```kotlin
fun quux() {
    try {
        foo()
        bar()
    } catch (_: Throwable) {
        foo()
        bar()
    }
}
```
```kotlin
fun quux() {
    try {
        foo()
        bar()
    } catch (_: Throwable) {}
}
```
```kotlin
fun quux() {
    runnnnnn {
        foo()
        bar()
    } + baz()
}
```
```kotlin
fun quux() {
    runnnnnn {
        foo()
        bar()
    } inny baz()
}
```
so I'd call this "precedent"

ptitjes:
This PR fixes the indentation of trailing lambda with chained calls.
See #589.

I used an AI agent to understand how `ktfmt` and `google-java-format` work. This PR has been implemented with the help of an AI agent. I don't know much about the PSI AST shapes, so I wouldn't be surprised if the AST tests and visits are not general enough, or simply wrong.

### DIFF
--- a/core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
+++ b/core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
@@ -9377,6 +9377,136 @@ class FormatterTest {
         .isEqualTo(expected)
   }
 
+  @Test
+  fun `correct indentation for lambda with chained function call (#589)`() =
+      assertFormatted(
+          """
+          |fun quux() = runnnnn {
+          |  foo()
+          |  bar()
+          |}
+          |    .baz()
+          |"""
+              .trimMargin()
+      )
+
+  @Test
+  fun `correct indentation for lambda with multiple chained function call (#589)`() =
+      assertFormatted(
+          """
+          |fun quux() = runnnnn {
+          |  foo()
+          |  bar()
+          |}
+          |    .baz {
+          |      foo()
+          |      bar()
+          |    }
+          |"""
+              .trimMargin()
+      )
+
+  @Test
+  fun `correct indentation for lambda with chained function call in block (#589)`() =
+      assertFormatted(
+          """
+          |fun quux() {
+          |  runnnnn {
+          |    foo()
+          |    bar()
+          |  }
+          |      .baz()
+          |}
+          |"""
+              .trimMargin()
+      )
+
+  @Test
+  fun `correct indentation for lambda with multiple chained function call in block (#589)`() =
+      assertFormatted(
+          """
+          |fun quux() {
+          |  runnnnn {
+          |    foo()
+          |    bar()
+          |  }
+          |      .baz {
+          |        foo()
+          |        bar()
+          |      }
+          |}
+          |"""
+              .trimMargin()
+      )
+
+  @Test
+  fun `correct indentation for lambda with multiple chained function call in nested block (#589)`() =
+      assertFormatted(
+          """
+          |fun quux() {
+          |  runnnnn {
+          |    foo()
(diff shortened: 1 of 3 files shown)


hick209:
This said no internal changes here. Which is odd, I was expecting a bunch of them.
Tomorrow I'll do some extra validation of it, but just wanted to let you know this much already

ptitjes:
@hick209 tell me if you need me to add more tests or fix something.

hick209:
Oops, look at me not keeping my promises. Sorry about that, will look at it again right now!

hick209 (changes requested):
Yes, this is incomplete as I suspected.

Here's the test I made, which I thought should happen the other way around
```
  @Test
  fun nivaldo_test() {
    // this passed
    assertFormatted(
        """
        |fun quux() {
        |  runnnnn {
        |        foo()
        |        bar()
        |      }
        |      .baz()
        |}
        |"""
            .trimMargin()
    )

    // this failed
    assertFormatted(
        """
        |fun quux() {
        |  runnnnn {
        |    foo()
        |    bar()
        |  }
        |      .baz()
        |}
        |"""
            .trimMargin()
    )
  }
```

Please LMK if I'm missing something here.

Also don't forget to update the `CHANGELOG.md` so I don't forget to give you proper attribution

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -9335,6 +9335,35 @@ class FormatterTest {
...
 
+  @Test
+  fun `correct indentation for lambda with chained function call (#589)`() =
+      assertFormatted(
+          """
+          |fun quux() = runnnnn {
+          |  foo()
+          |  bar()
+          |}
+          |  .baz()
```
hick209: Continuation indent is 4 for the style tested here.
I don't know if this is a bug on the new code or existing one though

ptitjes: Sorry. I fixed it to use the continuation indent and not the block indent.

ptitjes:
> Yes, this is incomplete as I suspected.

Hmmm, sorry, it seems I forgot half of the issue... I fixed it and added a bunch more tests.
Also fixed it to use the continuation indent for chained calls (and not the block indent as it was before).

> Also don't forget to update the `CHANGELOG.md` so I don't forget to give you proper attribution

Done.

I applied the new formatting (1 change in `KotlinInputAstVisitor.kt`) in a separate commit, so that you can see the changes. Now I expect you will have more changes on your internal projects...

hick209:
It still says no changes in our mega repo 🤔

Maybe we are missing something here. Let me test it out further to see if we are missing any use case

Edit: This might actually be a bug on our end here. I'm investigating
Edit 2: My script was too narrow and this pattern is not all that wide spread. That's why it was not caught. Seems like we are good now!

Thanks for the awesome work!

cortinico (approved):

PR #627, opened 2026-06-13
Add parser support for --lines, --offset, --length

### LINKED ISSUE #218: Support formatting of specific line ranges
The repeatable `--lines` option of google-java-format allows for formatting specific line ranges only rather than a whole file. This is useful for scripts/tools that want to run formatters only on lines that have been touched by a change, which avoids churn when formatter behavior changes.

An example of a tool that would utilize this is the [`fix`](https://www.mercurial-scm.org/repo/hg/file/stable/hgext/fix.py) Mercurial extension.

### LINKED ISSUE #573: Reformat before commit and partial reformat is not supported from IntelliJ plugin
Ktfmt intellij plugin does not account for line ranges nor declares `Feature.FORMAT_FRAGMENTS` in it's `KtfmtFormattingService` (see [FormattingServiceUtil.findService](https://github.com/JetBrains/intellij-community/blob/72c6c8e4ad498e8ae624b5f019063d79dbdc7acc/platform/code-style-impl/src/com/intellij/formatting/service/FormattingServiceUtil.java#L25)).

As a result, using "reformat before commit", formatting a selected portion of the code or other partial reformatting from the IDE result in formatting that do not comply with Ktfmt, leading to inconsistent states and need for manual reformatting of our repo every once in a while.

Git hooks or full reformat before commit are a workaround but is not convenient (breaks committing only a part of a file's changes to VCS) and requires extra setup and computation.

Reproduction sample:
- Committing with "Reformat code" commit check leads to the formatting below
- Reformatting with Ktfmt changes indents and puts the first `padding` in the same line as `Modifier`
```kotlin
private fun MyComposeFunction() {
    Function(
        modifier =
        Modifier
            .padding(vertical = someVerticalPadding())
            .padding(vertical = someVerticalPadding())
    )
}

fun Function(
    modifier: Any,
): Unit = TODO()

fun someVerticalPadding(): Any = TODO()

interface Modifier {
    fun padding(vertical: Any): Modifier = TODO()

    companion object : Modifier
}
```

ZacSweers:
Resolves #218 and includes IDE support as well.

Note that I opted to match GJF's current behavior and allow statement-level boundaries (the issue described function-level granularity only, but that's not what I see GJF do).

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
+++ b/core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
@@ -195,7 +195,10 @@ data class ParsedArgs(
                       nextValue()
                           ?: return ParseResult.Error("required value was not provided for: $key")
                     }
-                offsets.add(value.toIntOrNull() ?: return ParseResult.Error(invalidInt(key, value)))
+                offsets.add(
+                    value.toIntOrNull()
+                        ?: return ParseResult.Error("invalid integer value for $key: $value")
+                )
               }
           arg.startsWith("--length") ->
               arg.split('=', limit = 2).let { argSplit ->
@@ -210,7 +213,10 @@ data class ParsedArgs(
                       nextValue()
                           ?: return ParseResult.Error("required value was not provided for: $key")
                     }
-                lengths.add(value.toIntOrNull() ?: return ParseResult.Error(invalidInt(key, value)))
+                lengths.add(
+                    value.toIntOrNull()
+                        ?: return ParseResult.Error("invalid integer value for $key: $value")
+                )
               }
           arg.startsWith("--") -> return ParseResult.Error("Unexpected option: $arg")
           arg.startsWith("@") -> return ParseResult.Error("Unexpected option: $arg")
@@ -265,9 +271,6 @@ data class ParsedArgs(
       val parts = arg.split('=', limit = 2)
       return parts[1].takeIf { parts[0] == key || parts.size == 2 }
     }
-
-    private fun invalidInt(flag: String, value: String?): String =
-        "invalid integer value for $flag: $value"
 
     private fun parseLineRanges(
         lineRanges: RangeSet<Int>,
--- a/core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
+++ b/core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
@@ -94,8 +94,8 @@ data class ParsedArgs(
         |  --offset=<offset>                 Character offset to format, paired with --length.
         |                                        May be used multiple times.
         |  --length=<length>                 Character length to format, paired with --offset.
-        |                                        May be used multiple times. 0 formats the line
-        |                                        under the cursor.
+        |                                        May be used multiple times. 0 formats the whole
+        |                                        line containing the given --offset.
         |  --set-exit-if-changed             Sets exit code to 1 if any input file was not
         |                                        formatted/touched
         |  --do-not-remove-unused-imports    Leaves all imports in place, even if not used
--- a/core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
+++ b/core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
@@ -19,6 +19,9 @@ package com.facebook.ktfmt.cli
 import com.facebook.ktfmt.format.Formatter
 import com.facebook.ktfmt.format.FormattingOptions
 import com.facebook.ktfmt.util.Ktfmt
+import com.google.common.collect.Range
+import com.google.common.collect.RangeSet
+import com.google.common.collect.TreeRangeSet
 import java.io.File
 import java.nio.charset.StandardCharsets.UTF_8
 
@@ -40,6 +43,11 @@ data class ParsedArgs(
     /** Suppress all non-error output. */
     val quiet: Boolean,
 ) {
+  /** Zero-indexed line ranges to format, using closed-open bounds, e.g. [0, 3) and [6, 7). */
+  internal val lineRanges: RangeSet<Int> = TreeRangeSet.create()
+  /** Zero-indexed character ranges to format, using closed-open bounds. */
+  internal val characterRanges: RangeSet<Int> = TreeRangeSet.create()
+
   companion object {
 
     fun processArgs(args: Array<String>): ParseResult {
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
```diff
@@ -165,13 +248,65 @@ data class ParsedArgs(
...
+      parsedArgs.characterRanges.addAll(characterRanges)
+      return ParseResult.Ok(parsedArgs)
     }
 
     private fun parseKeyValueArg(key: String, arg: String): String? {
       val parts = arg.split('=', limit = 2)
       return parts[1].takeIf { parts[0] == key || parts.size == 2 }
     }
+
+    private fun String?.toIntOrNull(): Int? {
```
hick209: Any reason why we re-implement this here?
Isn't there an equivalent in the core Kotlin API?

core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
```diff
@@ -136,10 +164,51 @@ data class ParsedArgs(
...
+          arg.startsWith("--lines=") ->
+              when (val result = parseLineRanges(lineRanges, parseKeyValueArg("--lines", arg))) {
+                LineRangeParseResult.Success -> Unit
+                is LineRangeParseResult.Error -> return ParseResult.Error(result.message)
+              }
+          arg.startsWith("--line=") ->
+              when (val result = parseLineRanges(lineRanges, parseKeyValueArg("--line", arg))) {
+                LineRangeParseResult.Success -> Unit
+                is LineRangeParseResult.Error -> return ParseResult.Error(result.message)
+              }
```
hick209: Could likely be simplified to something like this or similar. Same goes for the options below
```
    arg.startsWith("--line") -> {
      val argSplit = arg.split('=', limit = 2)
      val key = argSplit.first()
      if (key != "--lines" && key != "--line") {
        ParseResult.Error("Unexpected option: $key")
      }
      val value = if (argSplit.size > 1) argSplit.last() else nextValue() ?: return ParseResult.Error("required value was not provided for: $key")

      when (val result = parseLineRanges(lineRanges, value)) {
        LineRangeParseResult.Success -> Unit
        is LineRangeParseResult.Error -> return ParseResult.Error(result.message)
      }
    }
```

core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
```diff
@@ -165,13 +248,65 @@ data class ParsedArgs(
...
+    private fun String?.toIntOrNull(): Int? {
+      return try {
+        this?.toInt()
+      } catch (_: NumberFormatException) {
+        null
+      }
+    }
+
+    private fun invalidInt(flag: String, value: String?): String =
+        "invalid integer value for $flag: $value"
```
hick209: Doesn't feel like this method adds much value. I won't stop if you like it though

core/src/test/java/com/facebook/ktfmt/cli/ParsedArgsTest.kt
```diff
@@ -121,6 +124,105 @@ class ParsedArgsTest {
     assertThat(parsed.stdinName).isEqualTo("my/foo.kt")
   }
 
+  @Test
+  fun `parseOptions recognizes --lines ranges`() {
+    val parsed =
+        assertSucceeds(ParsedArgs.parseOptions(arrayOf("--lines=1:3,5", "--lines", "7", "foo.kt")))
```
hick209: Maybe create a helper to reduce boilerplate here (not on you really, but since you are touching it, you know 😅)

```
// In this test file here
private fun parseOptions(vararg options: String) =
    ParsedArgs.parseOptions(options.toTypedArray())
```

core/src/test/java/com/facebook/ktfmt/cli/ParsedArgsTest.kt
```diff
@@ -271,6 +375,20 @@ class ParsedArgsTest {
...
+    parsedArgs.lineRanges.addAll(lineRanges)
+    parsedArgs.characterRanges.addAll(characterRanges)
+    return ParseResult.Ok(parsedArgs)
+  }
+
+  private fun lineRanges(vararg ranges: Range<Int>): RangeSet<Int> {
+    return ranges(*ranges)
+  }
+
+  private fun ranges(vararg ranges: Range<Int>): RangeSet<Int> {
```
hick209: Why both of those here? Let's just pick one and go with that

core/src/test/java/com/facebook/ktfmt/cli/ParsedArgsTest.kt
```diff
@@ -271,6 +375,20 @@ class ParsedArgsTest {
             editorConfig,
             quiet,
         )
-    )
+    parsedArgs.lineRanges.addAll(lineRanges)
+    parsedArgs.characterRanges.addAll(characterRanges)
```
hick209: The logic here doesn't feel like it should live here for me.
Thoughts?

hick209 (changes requested):
Just haven't finished `core/src/main/java/com/facebook/ktfmt/format/Formatter.kt`, but here's the start.

I'll resume tomorrow

core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
ZacSweers: I thought there was but couldn't find it in autocomplete for some reason, but now I see it 😅

ZacSweers: [9a32217](https://github.com/facebook/ktfmt/pull/627/commits/9a322171763ce08313bbb865d43baf5c0729dd11)

core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
ZacSweers: [bfb7ff2](https://github.com/facebook/ktfmt/pull/627/commits/bfb7ff26673231cc635ffa0b4c064d4499cafac9)

core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt
ZacSweers: [a7589d0](https://github.com/facebook/ktfmt/pull/627/commits/a7589d06e37e1d7d23cdc7052ae407d223307382)

core/src/test/java/com/facebook/ktfmt/cli/ParsedArgsTest.kt
ZacSweers: [becd1ac](https://github.com/facebook/ktfmt/pull/627/commits/becd1ac7eac000c0630ad23572d61670f138ec83)

core/src/test/java/com/facebook/ktfmt/cli/ParsedArgsTest.kt
ZacSweers: [e02a41a](https://github.com/facebook/ktfmt/pull/627/commits/e02a41ab9eaba4221ee66890fb7501e1eb36c538)

core/src/test/java/com/facebook/ktfmt/cli/ParsedArgsTest.kt
ZacSweers: [72e3e01](https://github.com/facebook/ktfmt/pull/627/commits/72e3e0148192834b65e67c08e0c87047dab41183)

PR #628, opened 2026-06-13
Switch in-repo formatting to simple local gradle task

ZacSweers:
The previous setup of using the ktfmt Gradle plugin but backdooring in the local ktfmt was finicky and resulting periodic stale runs and prone to ABI breakages. This switches the repo to just run a simple local JavaExec that consumes the compiler artifact directly. This way it's always aligned and doesn't get dinged by Gradle hacks.

Running this also caught and fixed a number of formatting issues on `main`!

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -35,10 +35,6 @@ version = providers.gradleProperty("ktfmt.version").get()
 
 tasks.wrapper { distributionType = Wrapper.DistributionType.ALL }
 
-repositories {
-  mavenCentral()
-}
-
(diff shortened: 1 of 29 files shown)


ZacSweers:
the codeql failure seems like a CI issue?

hick209:
Yeah, I need to fix this codeql thing, don't worry about it.

Great set of changes here. I sent you some comments on slack about it

ZacSweers:
Fixed the merge conflicts!

cortinico (approved):

PR #630, opened 2026-06-18
Support name-based destructuring declarations

### LINKED ISSUE #629: Missing support for name-based destructuring
Tested with ktfmt IDEA plugin 1.3.0.63.

```kotlin
data class D(val a: Int, val b: Int)
val d = D(1, 2)
```

## Successes

Save the file and see the spaces removed.

```kotlin
val (a, b         ) = d // experimental name-based destructuring syntax, which looks the same as position-based in stable kotlin
```

## Failures

Ktfmt is kind of stuck (IDEA shows "Running ktfmt..." but it doesn't complete) while formatting the file, which timeouts after 30 seconds on IDEA. The spaces are not removed.

```kotlin
val [a, b         ] = d // experimental position-based destructuring syntax
val (a, x = b         ) = d // experimental name-based destructuring syntax with renaming
```

## Build Script

```kotlin
// build.gradle.kts

kotlin {
  jvmToolchain(21)
  compilerOptions {
    jvmTarget = JvmTarget.JVM_21
    freeCompilerArgs.addAll(
      "-Xname-based-destructuring=complete",
    )
  }
}
```

goutamadwant:
## Summary

- Preserve square-bracket destructuring delimiters instead of always emitting parentheses.
- Format name-based destructuring entries with renames such as `x = b`.
- Add regression coverage for bracket destructuring and rename spacing.

Fixes #629

## Tests

- `./gradlew :ktfmt:test --tests "com.facebook.ktfmt.format.FormatterTest.handle name based destructuring declaration" --no-daemon --console=plain --stacktrace`
- `./gradlew :ktfmt:test --tests "com.facebook.ktfmt.format.FormatterTest" --no-daemon --console=plain --stacktrace`
- `./gradlew :ktfmt:build --no-daemon --console=plain --stacktrace`

Note: I also tried `./gradlew build --no-daemon --console=plain --stacktrace`; the local run reached ktfmt checks, then failed in unrelated `:lambda:test` dependency resolution because local Maven is missing `io.netty:netty-transport-native-epoll:4.1.42.Final:linux-x86_64`.

### DIFF
--- a/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
@@ -2248,7 +2248,7 @@ class KotlinInputAstVisitor(
     builder.token("]")
   }
 
-  /** Example `val (a, b: Int) = Pair(1, 2)` */
+  /** Example `val (a, b: Int) = Pair(1, 2)` or `val [a, b] = Pair(1, 2)` */
   override fun visitDestructuringDeclaration(destructuringDeclaration: KtDestructuringDeclaration) {
     builder.sync(destructuringDeclaration)
     val valOrVarKeyword = destructuringDeclaration.valOrVarKeyword
@@ -2257,8 +2257,10 @@ class KotlinInputAstVisitor(
       builder.space()
     }
     val hasTrailingComma = destructuringDeclaration.trailingComma != null
+    val (openingDelimiter, closingDelimiter) =
+        if (destructuringDeclaration.hasSquareBrackets()) "[" to "]" else "(" to ")"
     builder.block(ZERO) {
-      builder.token("(")
+      builder.token(openingDelimiter)
       builder.breakOp(Doc.FillMode.UNIFIED, "", expressionBreakIndent)
       builder.block(expressionBreakIndent) {
         visitEachCommaSeparated(
@@ -2268,7 +2270,7 @@ class KotlinInputAstVisitor(
         )
       }
     }
-    builder.token(")")
+    builder.token(closingDelimiter)
     val initializer = destructuringDeclaration.initializer
     if (initializer != null) {
       builder.space()
@@ -2282,13 +2284,13 @@ class KotlinInputAstVisitor(
     }
   }
 
-  /** Example `a: String` which is part of `(a: String, b: String)` */
+  /** Example `a: String` or `x = a` which is part of `(a: String, x = b)` */
   override fun visitDestructuringDeclarationEntry(
       multiDeclarationEntry: KtDestructuringDeclarationEntry
   ) {
     builder.sync(multiDeclarationEntry)
     declareOne(
-        initializer = null,
+        initializer = multiDeclarationEntry.initializer,
         kind = DeclarationKind.PARAMETER,
         modifiers = multiDeclarationEntry.modifierList,
         name = multiDeclarationEntry.nameIdentifier?.text ?: fail(),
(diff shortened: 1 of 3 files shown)


hick209:
This is terrific! Thank you for your contribution here. Don't forget to update the `CHANGELOG.md`.

I think we will still need to handle the case of `(val mail = email, val name = username) = user`, right?
But this PR as is is already an improvement, so I'd take as it is, just update the `CHANGELOG.md`.

hick209 (approved):

goutamadwant:
Thanks @hick209 updated CHANGELOG.md

hick209:
That was fast!

goutamadwant:
@hick209 I see `Facebook Internal - Linter` and `Facebook Internal - Builds & Tests` are failing behind internal links that I cannot access.
are these actionable for this PR? I can follow up if there is a code change/fix needed on my side.

core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -2282,13 +2284,13 @@ class KotlinInputAstVisitor(
     }
   }
 
-  /** Example `a: String` which is part of `(a: String, b: String)` */
+  /** Example `a: String` or `x = a` which is part of `(a: String, x = b)` */
```
cortinico:
```suggestion
  /** Example `a: String` or `x = a` which is part of `(a: String, x = a)` */
```

cortinico (approved):
LGTM but minor comment

hick209:
@goutamadwant we got this error here
```
core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt:2261:38: error: unresolved reference 'hasSquareBrackets'.
        if (destructuringDeclaration.hasSquareBrackets()) "[" to "]" else "(" to ")"
                                     ^^^^^^^^^^^^^^^^^
```

Internally we are still on Kotlin 2.2, which might be the reason for the issue. Do you know if there's an alternative here? Else we could work to make this backwards compatible for a while

goutamadwant:
Thanks @hick209 yes.. I think the compatibility path is to avoid `hasSquareBrackets()` and use `lPar` / `rPar` text instead.

From what I checked, `getLPar()` / `getRPar()` exist in Kotlin 2.2, and with Kotlin 2.3 they still resolve the bracket delimiters for this syntax. I can also guard the new regression test with `KotlinVersion.CURRENT < KotlinVersion(2, 3)` so internal Kotlin 2.2 builds do not try to parse the newer syntax.

I can push this follow up if this sounds good to you.

hick209:
Yep, sounds good!

goutamadwant:
addressed the issue @hick209.. it should compile against Kotlin 2.2 while still preserving bracket delimiters on Kotlin 2.3. Let me know if it looks good now. Thanks!

core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
```diff
@@ -3074,25 +3074,28 @@ class FormatterTest {
...
-              """
-              |fun f(d: D) {
-              |  val [a, b] = d
-              |  val (a, x = b) = d
-              |}
-              |"""
-                  .trimMargin()
-          )
+  fun `handle name based destructuring declaration`() {
+    if (KotlinVersion.CURRENT < KotlinVersion(2, 3)) return
```
hick209, line 3078: Is this check here necessary?
I feel like it would pass regardless

hick209 (approved):

Issue #631, opened 2026-06-23
Brackets are formatted in a weird way when followed by chained calls


Yqloss:
Current behavior:

```kotlin
fun main() {
  {
      println("-----------------------------------")
      println("-----------------------------------")
    }
    .invoke()

  [
      println("-----------------------------------"),
      println("-----------------------------------"),
    ]
    .map { it }

  val a =
    [
        println("-----------------------------------"),
        println("-----------------------------------"),
      ]
      .map { it }
}
```

They just feel weird, and this problem would get worse and worse with the stabilization of collection literals because the pattern of constructing a list and do operations on it is so common. The current formatting is too weird to be acceptable.

Why don't we put ] ) } at the same indentation of the line containing [ ( {, and move the dot to align with ] ) } as well, so that everything would look nice?

```kotlin
fun main() {
  {
    println("-----------------------------------")
    println("-----------------------------------")
  }
  .invoke()

  [
    println("-----------------------------------"),
    println("-----------------------------------"),
  ]
  
  [
    println("-----------------------------------"),
    println("-----------------------------------"),
  ]
  .map { it }

  val a =
    [
      println("-----------------------------------"),
      println("-----------------------------------"),
    ]
    .map { it }

  run(
    block = {
      println("-----------------------------------")
      println("-----------------------------------")
    }
  )
}
```

hick209:
Thanks for flagging this.
How about this? It would align with what we do with lambdas
```
fun main() {
  {
    println("-----------------------------------")
    println("-----------------------------------")
  }
      .invoke()

  [
    println("-----------------------------------"),
    println("-----------------------------------"),
  ]
  
  [
    println("-----------------------------------"),
    println("-----------------------------------"),
  ]
      .map { it }

  val a = [
      println("-----------------------------------"),
      println("-----------------------------------"),
  ]
      .map { it }

  run(
    block = {
      println("-----------------------------------")
      println("-----------------------------------")
    }
  )
}
```

Yqloss:
I think that's a pretty balanced solution. Looks good.

By the way, in the 4th example, ktfmt would put the [] on the next line:

```kotlin
fun main() {
  val a =
    [
      println("-----------------------------------"),
      println("-----------------------------------"),
    ]
      .map { it }
}
```

ptitjes:
@hick209 can I pick this?

hick209:
@ptitjes go for it!

PR #634, opened 2026-06-25
Indentation of block-like calls with or without chained call

### LINKED ISSUE #589: Strange formatting for lambda with chained function call

### LINKED ISSUE #633: Non-matching indent on method call followed by chained call
In the vein of https://github.com/facebook/ktfmt/issues/589 and https://github.com/facebook/ktfmt/issues/631, I just noticed that, the formatting of method calls with chained calls is also strange/weird.

`ktfmt` considers the following formatted:
```kotlin
fun quux() {
  listOf(
      1 to "one",
      2 to "two",
  )

  listOf(
          1 to "one",
          2 to "two",
      )
      .forEach { (left, right) ->
        foo()
        bar()
      }
}
```

The method call and its closing parenthesis are not on the same column when the call is followed by a chained call.

I believe that the correct formatting should be:
```kotlin
fun quux() {
  listOf(
      1 to "one",
      2 to "two",
  )

  listOf(
      1 to "one",
      2 to "two",
  )
      .forEach { (left, right) ->
        foo()
        bar()
      }
}
```

for at least two reasons:
- obviously, the indent levels of the parenthesis would match, and the indent level of the arguments would not change.
- adding a chained call to a prior method call, would not modify the formatting of that method call. This makes the formatting more stable and vcs-friendly.

## Reproduction

Both the tests bellow are passing. I think only the first one should.

```kotlin
    @Test
    fun `correct indentation for call without chained function call`() =
        assertFormatted(
            """
              |fun quux() {
              |  listOf(
              |      1 to "one",
              |      2 to "two",
              |  )
              |}
              |"""
                .trimMargin()
        )

    @Test
    fun `correct indentation for call with multiple chained function call`() =
        assertThatFormatting(
            """
              |fun quux() {
              |  listOf(
              |      1 to "one",
              |      2 to "two",
              |  )
              |      .forEach { (left, right) ->
              |        foo()
              |        bar()
              |      }
              |}
              |"""
                .trimMargin()
        )
            .isEqualTo(
                """
              |fun quux() {
              |  listOf(
              |          1 to "one",
              |          2 to "two",
              |      )
              |      .forEach { (left, right) ->
              |        foo()
              |        bar()
              |      }
              |}
              |"""
                    .trimMargin()
            )
```

ptitjes:
Closes #633

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -75,11 +75,10 @@ tasks {
 
   // Javadoc
   register("javadocJar", Jar::class) {
-    val dokkaJavadocTask =
-        named(
-            "dokkaGeneratePublicationJavadoc",
-            org.jetbrains.dokka.gradle.tasks.DokkaGeneratePublicationTask::class,
-        )
+    val dokkaJavadocTask = named(
+        "dokkaGeneratePublicationJavadoc",
+        org.jetbrains.dokka.gradle.tasks.DokkaGeneratePublicationTask::class,
+    )
     dependsOn(dokkaJavadocTask)
     from(dokkaJavadocTask.flatMap { it.outputDirectory })
     archiveClassifier.set("javadoc")
--- a/core/src/main/java/com/facebook/ktfmt/format/EnumEntryList.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/EnumEntryList.kt
@@ -60,13 +60,12 @@ private constructor(
 
       var semicolon: PsiElement? = null
       var comma: PsiElement? = null
-      val lastToken =
-          checkNotNull(
-              enumEntries
-                  .last()
-                  .lastChild
-                  .getPrevSiblingIgnoringWhitespaceAndComments(withItself = true),
-          )
+      val lastToken = checkNotNull(
+          enumEntries
+              .last()
+              .lastChild
+              .getPrevSiblingIgnoringWhitespaceAndComments(withItself = true),
+      )
       when (lastToken.text) {
         "," -> {
           comma = lastToken
--- a/core/src/main/java/com/facebook/ktfmt/format/Formatter.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/Formatter.kt
@@ -44,27 +44,24 @@ import org.jetbrains.kotlin.psi.psiUtil.startOffset
 object Formatter {
 
   @JvmField
-  val META_FORMAT =
-      FormattingOptions(
-          blockIndent = 2,
-          continuationIndent = 4,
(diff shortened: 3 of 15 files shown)


Yqloss:
```kotlin
  [
      println("-----------------------------------"),
      println("-----------------------------------"),
    ]
    .map { it }
```

isn't fixed in this pr. In other words, this needs to be more general, or to be applied on more types of elements.

Edit:

The title of this pr is also too narrow. Not only calls but all kinds of brackets have this issue. They all need to be fixed or we would get an inconsistent behavior and weird formatting.

ptitjes:
@Yqloss I will address https://github.com/facebook/ktfmt/issues/631 in a separate PR.

@hick209 this change is very aggressive, as you can see by the third commit that applies the new formatting to the code-base. I made the block-like calls start on the same line for function body expressions and for property delegate/initializer. This makes for a more compact output and more consistent with how blocks and scoping lambdas are formatted. This however will produce a lot of changes in every code-bases...

hick209:
@ptitjes, that's somewhat expected and part of the reason why I initially didn't want to go into this rabbit hole.
Well, we're in the middle of it now, so I guess the best way out if through.

If you prefer, we can make one tweak at a time and land them as multiple PRs.
Up to you here though

ptitjes:
> If you prefer, we can make one tweak at a time and land them as multiple PRs.
> Up to you here though

Hey @hick209. I am not sure I understood you... Do you want me to split this PR?
If it is good enough for you as is, I am OK with not splitting it... :smile: 

I don't think my implementation is the most efficient but maybe we can optimize later.

cortinico (approved):

PR #635, opened 2026-06-27
Indentation of block-like brackets followed by chained call

### LINKED ISSUE #589: Strange formatting for lambda with chained function call

### LINKED ISSUE #631: Brackets are formatted in a weird way when followed by chained calls

ptitjes:
Closes #631

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/ktfmt/format/KotlinInputAstVisitor.kt
+++ b/core/src/main/kotlin/org/jetbrains/ktfmt/format/KotlinInputAstVisitor.kt
@@ -1708,24 +1708,31 @@ open class KotlinInputAstVisitor(
    * operator (instead of breaking and indenting after it), and any chained selectors break onto
    * their own line, mirroring how scoping functions and lambdas are handled.
    */
-  @OptIn(ExperimentalContracts::class)
   private fun isBlockLikeCall(expression: KtExpression?): Boolean {
-    contract { returns(true) implies (expression is KtCallExpression) }
-
     if (expression == null) return false
     val prev = expression.getPrevSiblingIgnoringWhitespace()
     if (prev is PsiComment) {
       return false // Leading comments cause weird indentation; keep the default layout.
     }
 
-    if (expression !is KtCallExpression) return false
-    val valueArgumentList = expression.valueArgumentList ?: return false
-    if (valueArgumentList.trailingComma != null) return true
-    return valueArgumentList.arguments.any { argument ->
-      val argumentExpression = argument.getArgumentExpression()
-      argumentExpression != null &&
-          (isBlockLikeCall(argumentExpression) || isChainedBlockLikeCall(argumentExpression))
+    if (expression is KtCollectionLiteralExpression) {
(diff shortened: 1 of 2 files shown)


hick209:
I think we need a rebase of this one here, it failed to import

ptitjes:
@hick209 I am really sorry I did not have the chance to work on this in the last weeks. I intend to be back at it during August. I hope this can wait then.

ptitjes:
@hick209 I fully rebased the PR on main, cleaned up the implementation, fixed an existing test and added an additional test.

Disclaimer: the tests and the initial implementation were LLM-generated. I reviewed and modified the implementation manually.

AbdullinAM:
@ptitjes Hey! We're currently in the process of changing a lot of stuff around, especially with tests. So this will probably require another rebase soon. I can take it over and rebase/merge this when everything will stabilise if there are no other comments from @hick209

hick209:
Yeah, no extra comments from my end

PR #636, opened 2026-06-28
Fix non-idempotent formatting when an auto-inserted trailing comma overflows max_width

### LINKED ISSUE #472: Google style with trailing comma does not maxWidth correctly

MathieuLorber:
## Summary

Formatting was not idempotent: running ktfmt twice on the same file could produce a different result than running it once.

Trailing commas are inserted as a text edit **after** the line-breaking engine has run. When an argument/parameter list only becomes multi-line **during** layout, the trailing comma that ktfmt then adds was never seen by the engine. If the last element lands exactly on `max_width`, the inserted comma pushes the line to `max_width + 1` — over the limit. On the next run the comma is now real input, so the engine breaks that line differently. Hence the instability.

This builds on #610, which added trailing commas to lists that were *already* multi-line in the source (before layout); this change covers lists that only become multi-line *during* layout.

## Reproduction

Single-line input (the second argument is sized so it sits exactly at the 100-column limit):

```kotlin
functionCall(shortArg, namedArgument = "string sized so the trailing comma ktfmt adds tips it one char over the limit.")
```

`--kotlinlang-style`.

**Before this fix** — pass 1 leaves the argument on one line, one char over the limit, and a
second pass is needed to break it (non-idempotent):

```kotlin
functionCall(
    shortArg,
    namedArgument = "string sized so the trailing comma ktfmt adds tips it one char over the limit.",
)
```

**After** — the value breaks onto its own line in a single pass:

```kotlin
functionCall(
    shortArg,
    namedArgument =
        "string sized so the trailing comma ktfmt adds tips it one char over the limit.",
)
```

Standalone reproducible repo: https://github.com/MathieuLorber/ktfmt-63-non-idempotent-example (initially noticed the problem with 0.63, but still true with 0.64)

## Fix

After the layout pass, re-run the layout while trailing commas keep being added, so the inserted commas are accounted for. Properties:

- No-op extra pass for already-formatted code and when trailing-comma management is disabled (`addRedundantElements` returns the input unchanged, so it stops after a single layout pass).
- Converges, because comma insertion is monotonic (commas are only ever added, and a list with a trailing comma is forced multi-line).

## Tests

Added 2 regression tests (kotlinlang style) for the case above.

## Note for reviewers — implementation style

This PR expresses the fixed-point loop as a `private tailrec fun`. If you'd rather keep the imperative style of the surrounding code, an identical fix using a `while` loop is in branch [`fix-idempotence-after-trailing-comma-with-while`](https://github.com/MathieuLorber/ktfmt/blob/a40b2d0cc8cec4156be75d5f664a187797316ded/core/src/main/java/com/facebook/ktfmt/format/Formatter.kt#L117) — happy to swap to it.
For variable naming and the == comparison I followed FormatterContext's style. 

Thanks !

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ### Changed
 
+### Fixed
+
+* Fix non-idempotent formatting when a managed trailing comma pushes a line over MAX_WIDTH (e.g. a long qualified expression as the last argument of a call). The comma is now accounted for by re-running the layout, so the line is broken correctly on the first pass. (https://github.com/facebook/ktfmt/pull/636)
+
 
 ## [0.64]
 
--- a/core/src/main/java/com/facebook/ktfmt/format/Formatter.kt
+++ b/core/src/main/java/com/facebook/ktfmt/format/Formatter.kt
@@ -124,8 +124,7 @@ object Formatter {
               .transform { sortedAndDistinctImports(it) }
               .transform { dropRedundantElements(it, options) }
               .transform { addRedundantElements(it, options) }
-              .transform { prettyPrint(it, options, lineSeparator = "\n") }
-              .transform { addRedundantElements(it, options) }
+              .let { prettyPrintAndManageTrailingCommas(it, options, lineSeparator = "\n") }
               .transform { MultilineStringFormatter(options.continuationIndent).format(it) }
               .code
         } else {
@@ -162,6 +161,21 @@ object Formatter {
     return formattedCode
         .let { convertLineSeparators(it, checkNotNull(Newlines.guessLineSeparator(kotlinCode))) }
         .let { if (shebang.isEmpty()) it else shebang + "\n" + it }
+  }
+
+  /**
+   * Pretty-prints & reprints while [addRedundantElements] keeps adding trailing commas, so a comma
+   * inserted after layout can't leave a line over the limit.
+   */
+  private tailrec fun prettyPrintAndManageTrailingCommas(
+      context: FormatterContext,
+      options: FormattingOptions,
+      lineSeparator: String,
+  ): FormatterContext {
+    val prettyCode = context.transform { prettyPrint(it, options, lineSeparator) }
+    val newCode = prettyCode.transform { addRedundantElements(it, options) }
+    return if (newCode == prettyCode) prettyCode
+    else prettyPrintAndManageTrailingCommas(newCode, options, lineSeparator)
   }
 
   /** prettyPrint reflows 'code' using google-java-format's engine. */
--- a/core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
+++ b/core/src/test/java/com/facebook/ktfmt/format/FormatterTest.kt
@@ -9179,6 +9179,41 @@ class FormatterTest {
   }
 
   @Test
+  fun `a trailing comma added while breaking a list does not push the last line over the max width`() {
+    val code =
+        """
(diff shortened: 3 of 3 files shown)


hick209:
Merge conflicts

MathieuLorber:
I've resolved the conflict with main. I have one note on scope: this fixes the idempotency issue on the whole-file path. The partial-formatting path added in #627 (--lines / --offset / --length) has its own copy of the cleanup pipeline (dropRedundantElements → sortedAndDistinctImports → addRedundantElements), where the same pattern applies, the trailing comma is inserted after layout and isn't re-measured.

I tried to naively apply the same fix, but it doesn't transfer directly: prettyPrintAndManageTrailingCommas re-runs a full-file prettyPrint, which would reformat code outside the selected ranges and breaks the --lines/--offset cases in MainTest. Doing it properly there would mean re-laying out only the affected ranges.

Since the two pipelines in format() share most of their steps but aren't unified, a fix on one doesn't automatically carry over to the other (the behavioral tests in FormatterTest run the whole-file path; the partial path is covered by a few MainTest integration tests).

(btw, any preference between the tailrec and while implementation ?)

MathieuLorber:
Exploring a follow-up: routing both whole-file and partial formatting through one range-aware trailing-comma pass (also fixes the partial path's max-width overflow). WIP here, tests green: https://github.com/MathieuLorber/ktfmt/pull/1/files

cortinico (approved):

hick209:
Regarding preference between the tailrec and while implementation, I'd say to go with the simplest one to avoid unnecessary complexity at this point.
If we ever measure perf and this is a problem, we can then improve it. Until then, let's keep it simple.

Now regarding the partial format thing, indeed that one is tricky and there's some sad duplication there.
Feel free to try to unify it if you can, but not required for this PR.

hick209:
I'll ship this PR as is, as it's an improvement.
If desired you can push another PR that adds this behavior to the partial formatting. How does that sound @MathieuLorber ?

MathieuLorber:
Sounds perfect, thank you :)

PR #639, opened 2026-07-01
Remove effectively dead code for tabular alignment

qwwdfsad:
* getPositionToColumnMap is a JavaInput API used for multidimensional array initializers alignment
* It was never computed properly because KotlinTok.column is always 0


core/api/ktfmt.api
```diff
@@ -271,7 +271,7 @@ public final class com/facebook/ktfmt/format/KotlinInputAstVisitor : org/jetbrai
 }
 
 public final class com/facebook/ktfmt/format/KotlinTok : com/google/googlejavaformat/Input$Tok {
-	public fun <init> (ILjava/lang/String;Ljava/lang/String;IIZLorg/jetbrains/kotlin/lexer/KtToken;)V
+	public fun <init> (ILjava/lang/String;Ljava/lang/String;IZLorg/jetbrains/kotlin/lexer/KtToken;)V
```
qwwdfsad, line 274: @hick209 could you please elaborate on whether ktfmt, as a backwards-compatible programmatic dependency, is something we care about?

So far, I don't have a mental model to follow -- e.g. in #552 it was okay not to preserve signatures

hick209: Basically the only thing I am careful about is public APIs that I'm aware are being used.
Like the Gradle and Spotless plugins, which I learned about after breaking it.

I don't see backwards-compat a bad thing to invest in, but won't block changes on not having it either unless I know the change is riskier, which right now it's mostly things around the formatting options.

hick209:
I mean, I'm okay with pulling this, but any issues with leaving it in?

qwwdfsad:
Not much issues. I've spotted it in a profiler (as something insignificant), have taken a look, was surprised that the column was always 0, dug a bit deeper, and here we are. So it's effectively dead and incorrect (because of 0) code that still does some computation at the runtime

PR #641, opened 2026-07-14
Sort modifiers according to Kotlin conventions

### LINKED ISSUE #293: Sort modifiers based on Kotlin conventions
When modifiers aren't in a specific order, the linter suggests to be sorted, ktfmt could sort them.

user1223644:
Sorts declaration modifiers according to the order defined by the Kotlin coding conventions.

For example:

```kotlin
final @Magic public class Foo

suspend override protected fun foo() {}
```

is formatted as:

```kotlin
@Magic public final class Foo

protected override suspend fun foo() {}
```

Annotations and comments keep their relative association. Unsupported modifiers such as `noinline`, `crossinline`, `in`, `out`, and `reified` act as ordering barriers.

The cleanup applies to both full and partial formatting.

Tests:

```text
./gradlew ktfmtFormat check --no-daemon
```

Fixes #293

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/ktfmt/format/ModifierSorter.kt
+++ b/core/src/main/kotlin/org/jetbrains/ktfmt/format/ModifierSorter.kt
@@ -1,0 +1,210 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jetbrains.ktfmt.format
+
+import org.jetbrains.kotlin.com.intellij.psi.PsiComment
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
+import org.jetbrains.kotlin.com.intellij.psi.PsiWhiteSpace
+import org.jetbrains.kotlin.lexer.KtModifierKeywordToken
+import org.jetbrains.kotlin.psi.KtAnnotation
+import org.jetbrains.kotlin.psi.KtAnnotationEntry
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtModifierList
+import org.jetbrains.kotlin.psi.KtTreeVisitorVoid
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
+import org.jetbrains.kotlin.psi.psiUtil.startOffset
+
+/** Sorts declaration modifiers according to the Kotlin coding conventions. */
+internal object ModifierSorter {
+  private val modifierRanks = listOf(
+      setOf("public", "protected", "private", "internal"),
+      setOf("expect", "actual"),
+      setOf("final", "open", "abstract", "sealed"),
+      setOf("const"),
+      setOf("external"),
+      setOf("override"),
+      setOf("lateinit"),
+      setOf("tailrec"),
+      setOf("vararg"),
+      setOf("suspend"),
+      setOf("inner"),
+      setOf("enum", "annotation", "fun"),
+      setOf("companion"),
+      setOf("inline", "value"),
+      setOf("infix"),
+      setOf("operator"),
+      setOf("data"),
+  )
+      .flatMapIndexed { rank, modifiers -> modifiers.map { it to rank } }
+      .toMap()
+
+  internal fun sort(file: KtFile): String {
+    val replacements = mutableListOf<Replacement>()
+    file.accept(
+        object : KtTreeVisitorVoid() {
+          override fun visitModifierList(list: KtModifierList) {
+            sortedText(list)?.let { replacements.add(Replacement(list, it)) }
+            super.visitModifierList(list)
+          }
+        },
+    )
+
+    if (replacements.isEmpty()) return file.text
+    val innermostReplacements = replacements.filter { candidate ->
+      replacements.none { other ->
+        other !== candidate &&
+            other.element.startOffset >= candidate.element.startOffset &&
+            other.element.endOffset <= candidate.element.endOffset
+      }
+    }
+    val result = StringBuilder(file.text)
+    for (replacement in innermostReplacements.sortedByDescending { it.element.endOffset }) {
+      result.replace(
+          replacement.element.startOffset,
+          replacement.element.endOffset,
+          replacement.text,
+      )
+    }
+    val sortedCode = result.toString()
+    return if (innermostReplacements.size == replacements.size) sortedCode
+    else sort(Parser.parse(sortedCode))
+  }
+
+  private fun sortedText(list: KtModifierList): String? {
+    val significantChildren =
+        generateSequence(list.node.firstChildNode) { it.treeNext }
+            .map { it.psi }
+            .filterNot { it is PsiWhiteSpace }
+            .toList()
+    val parts = mutableListOf<Part>()
+    val sortableSegment = mutableListOf<PsiElement>()
+
+    fun flushSortableSegment() {
+      if (sortableSegment.isEmpty()) return
+      parts.add(Part.Sortable(sortSegment(list, sortableSegment)))
+      sortableSegment.clear()
+    }
+
+    for (child in significantChildren) {
+      if (child is PsiComment || child.isAnnotation() || child.sortRank() != null) {
+        sortableSegment.add(child)
+      } else {
+        flushSortableSegment()
+        parts.add(Part.Fixed(child.text))
+      }
+    }
+    flushSortableSegment()
+
+    val sorted = parts.map { it.text }.joinWithSpaces()
+    return sorted.takeIf { it != list.text }
+  }
+
+  private fun sortSegment(list: KtModifierList, elements: List<PsiElement>): String {
+    val units = mutableListOf<SortableUnit>()
+    val leadingComments = mutableListOf<PsiComment>()
+
+    for (element in elements) {
+      if (element is PsiComment) {
+        val previous = units.lastOrNull()
+        if (previous != null && list.hasNoNewlineBetween(previous.base, element)) {
+          previous.trailingComments.add(element)
+        } else {
+          leadingComments.add(element)
+        }
+        continue
+      }
+
+      units.add(
+          SortableUnit(
+              base = element,
+              rank = element.sortRank(),
+              isAnnotation = element.isAnnotation(),
+              leadingComments = leadingComments.toMutableList(),
+          ),
+      )
+      leadingComments.clear()
+    }
+
+    if (leadingComments.isNotEmpty()) {
+      units.lastOrNull()?.followingComments?.addAll(leadingComments)
+    }
+
+    val sortedUnits =
+        units.filter { it.isAnnotation } + units.filterNot { it.isAnnotation }.sortedBy { it.rank }
+    if (sortedUnits.map { it.base } == units.map { it.base }) {
+      return list.text.substring(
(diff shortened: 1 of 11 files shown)


cgrushko:
Thanks for the pull request!

For transparency, we'd appreciate if you can note whether / how much you've used the help of AI coding tools for creating this PR. Thanks!

user1223644:
I used codex extensively to help implement this change. I personally reviewed and tested all changes before submitting

hick209 (changes requested):
LGTM, but now it's up to Kotlin Foundations on whether they want to take this in or not.

There were also plenty of changes recently, so you will need a rebase.

core/src/main/kotlin/org/jetbrains/ktfmt/format/ModifierSorter.kt
```diff
@@ -0,0 +1,210 @@
...
+  private val modifierRanks = listOf(
+      setOf("public", "protected", "private", "internal"),
+      setOf("expect", "actual"),
+      setOf("final", "open", "abstract", "sealed"),
+      setOf("const"),
+      setOf("external"),
+      setOf("override"),
+      setOf("lateinit"),
+      setOf("tailrec"),
+      setOf("vararg"),
```
zarechenskiy, line 42: `vararg` is a parameter modifier. It seems we don't need it, do we?

Or, if we also want to tackle _all_ modifiers, then it makes sense to include noinline and other parameter modifiers as well

core/src/test/kotlin/org/jetbrains/ktfmt/format/ModifierSorterTest.kt
```diff
@@ -0,0 +1,197 @@
...
+            code,
+            lineRanges = selectedSecondLine,
+        ),
+    )
+  }
+
+  @Test
+  fun `modifier sorting preserves line separators and context receivers`() {
+    val code = "context(Something)\r\noverride public fun f() {}\r\n"
+    val expected = "context(Something)\r\npublic override fun f() {}\r\n"
```
zarechenskiy, line 193: Context receivers were replaced with context parameters, so let's use them in tests

core/src/test/kotlin/org/jetbrains/ktfmt/format/ModifierSorterTest.kt
```diff
@@ -0,0 +1,197 @@
...
+        Formatter.format(
+            Formatter.KOTLINLANG_FORMAT,
+            code,
+            lineRanges = selectedSecondLine,
+        ),
+    )
+  }
+
+  @Test
+  fun `modifier sorting preserves line separators and context receivers`() {
```
zarechenskiy, line 191: Speaking of context parameters: I believe we should put `context` to the first place among the modifiers. So while this test makes sense, we also have to check that something like:
```
internal context(other: Int)
fun something(s: String) {}
```

is formatted into:
```
context(other: Int)
internal fun something(s: String) {}
```

otherwise, there might be some confusion whether some modifier is applied to `context` or to a function (it always applies to a function)

core/src/main/kotlin/org/jetbrains/ktfmt/format/ModifierSorter.kt
```diff
@@ -0,0 +1,210 @@
...
+import org.jetbrains.kotlin.lexer.KtModifierKeywordToken
+import org.jetbrains.kotlin.psi.KtAnnotation
+import org.jetbrains.kotlin.psi.KtAnnotationEntry
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtModifierList
+import org.jetbrains.kotlin.psi.KtTreeVisitorVoid
+import org.jetbrains.kotlin.psi.psiUtil.endOffset
+import org.jetbrains.kotlin.psi.psiUtil.startOffset
+
+/** Sorts declaration modifiers according to the Kotlin coding conventions. */
```
zarechenskiy, line 31: This comment doesn't add much value, let's remove it

core/src/test/resources/cases/new_codestyle/annotation/DeclarationAnnotations.output
```diff
@@ -8,11 +8,11 @@
...
 @A private val beforeModifier = 1
 
-private @Deprecated("some reason") val afterModifier = 1
+@Deprecated("some reason") private val afterModifier = 1
 
-public @Magic final class Interleaved
+@Magic public final class Interleaved
 
-public @Magic(1, "argument") final class InterleavedWithArgs
+@Magic(1, "argument") public final class InterleavedWithArgs
```
zarechenskiy, line 15: Let's also check that we don't touch annotations in case of use-site targets, for instance:
```
@Target(AnnotationTarget.VALUE_PARAMETER)
annotation class Anno

class F {
    final internal val @receiver:Anno Int.prop get() = 4
}
```

Should be formatted as `internal final val @receiver:Anno Int.prop get() = 4`

zarechenskiy (changes requested):
I haven't really looked into the code, only at the expect/actual results. For the code part, @AbdullinAM will review it

core/src/test/kotlin/org/jetbrains/ktfmt/format/ModifierSorterTest.kt
```diff
@@ -0,0 +1,197 @@
...
+ */
+
+package org.jetbrains.ktfmt.format
+
+import com.google.common.collect.Range
+import com.google.common.collect.TreeRangeSet
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+class ModifierSorterTest {
```
AbdullinAM, line 24: I think we should stick to a new file-based tests format, see `FormatTest` for an example.

core/src/main/kotlin/org/jetbrains/ktfmt/format/ModifierSorter.kt
AbdullinAM: Yes, I think current implementation requires having all modifiers. Currently implementation fails with a parsing error on this example:
```
inline fun f(@Magic // c
noinline b: () -> Unit) {}
```

Modifier sorter does not consider `noinline` as modifier and removes the line break after comment, transforming code into invalid
```
inline fun f(@Magic // c noinline b: () -> Unit) {}
```

core/src/main/kotlin/org/jetbrains/ktfmt/format/ModifierSorter.kt
```diff
@@ -0,0 +1,210 @@
...
+  private data class SortableUnit(
+      val base: PsiElement,
+      val rank: Int?,
+      val isAnnotation: Boolean,
+      val leadingComments: MutableList<PsiComment>,
+      val trailingComments: MutableList<PsiComment> = mutableListOf(),
+      val followingComments: MutableList<PsiComment> = mutableListOf(),
+  )
+
+  private sealed interface Part {
```
AbdullinAM, line 203: It feels like this abstraction is not necessary. Both subclasses are only created in one place in code and no distinction is made between them after:
```
val parts = mutableListOf<Part>() // line 93
parts.add(...) // lines 98, 107
val sorted = parts.map { it.text }.joinWithSpaces() // line 112
```

Lets remove this abstraction. Or add documentation explaining why its necessary

core/src/main/kotlin/org/jetbrains/ktfmt/format/ModifierSorter.kt
```diff
@@ -0,0 +1,210 @@
...
+        sortableSegment.add(child)
+      } else {
+        flushSortableSegment()
+        parts.add(Part.Fixed(child.text))
+      }
+    }
+    flushSortableSegment()
+
+    val sorted = parts.map { it.text }.joinWithSpaces()
+    return sorted.takeIf { it != list.text }
```
AbdullinAM, line 113: One potential problem of this implementation is that we don't preserve the line breaks between modifiers. So we will trigger replacement/parsing even if the original code just contained line breaks like
```
public
override
fun outsideSelection() {}
```

AbdullinAM: I would suggest modifying the code to detect if the modifiers were actually sorted

core/src/main/kotlin/org/jetbrains/ktfmt/format/ModifierSorter.kt
```diff
@@ -0,0 +1,210 @@
...
+              rank = element.sortRank(),
+              isAnnotation = element.isAnnotation(),
+              leadingComments = leadingComments.toMutableList(),
+          ),
+      )
+      leadingComments.clear()
+    }
+
+    if (leadingComments.isNotEmpty()) {
+      units.lastOrNull()?.followingComments?.addAll(leadingComments)
```
AbdullinAM, line 143: To be honest I was not able to come up with a test case that covers this condition (except for test cases with currently unsupported modifiers). If we handle all modifiers properly --- would we even need that?

core/src/main/kotlin/org/jetbrains/ktfmt/format/ModifierSorter.kt
```diff
@@ -0,0 +1,210 @@
...
+    for (replacement in innermostReplacements.sortedByDescending { it.element.endOffset }) {
+      result.replace(
+          replacement.element.startOffset,
+          replacement.element.endOffset,
+          replacement.text,
+      )
+    }
+    val sortedCode = result.toString()
+    return if (innermostReplacements.size == replacements.size) sortedCode
+    else sort(Parser.parse(sortedCode))
```
AbdullinAM, line 84: I suspect that this sorter can be implemented in one pass (i.e. requiring only one re-parsing). I'm not 100% sure if we need it though, because it can complicate the code. So we can merge this version and improve it later

core/src/main/kotlin/org/jetbrains/ktfmt/format/ModifierSorter.kt
```diff
@@ -0,0 +1,210 @@
...
+    if (sortedUnits.map { it.base } == units.map { it.base }) {
+      return list.text.substring(
+          elements.first().startOffset - list.startOffset,
+          elements.last().endOffset - list.startOffset,
+      )
+    }
+    return sortedUnits.map { it.render() }.joinWithSpaces()
+  }
+
+  private fun PsiElement.isAnnotation(): Boolean = this is KtAnnotation || this is KtAnnotationEntry
```
AbdullinAM, line 157: We can give annotations rank `-1`, so we don't have to always handle them separately

core/src/main/kotlin/org/jetbrains/ktfmt/format/ModifierSorter.kt
```diff
@@ -0,0 +1,210 @@
...
+      )
+      leadingComments.clear()
+    }
+
+    if (leadingComments.isNotEmpty()) {
+      units.lastOrNull()?.followingComments?.addAll(leadingComments)
+    }
+
+    val sortedUnits =
+        units.filter { it.isAnnotation } + units.filterNot { it.isAnnotation }.sortedBy { it.rank }
```
AbdullinAM, line 147: Having a rank for annotations will simplify this code

core/src/main/kotlin/org/jetbrains/ktfmt/format/ModifierSorter.kt
```diff
@@ -0,0 +1,210 @@
...
+      if (comment.text.startsWith("//")) append('\n')
+    }
+    for (comment in followingComments) {
+      append('\n')
+      append(comment.text)
+      append('\n')
+    }
+  }
+
+  private fun Iterable<String>.joinWithSpaces(): String = buildString {
```
AbdullinAM, line 185: There are several cases in code where you use `list.map { ... }.joinWithSpaces()`. I would suggest you modify this fun to take a lambda that transforms the list element into string and then you can write:
```
list.joinWithSpaces { ... }
```

This will be similar to `joinToString` from stdlib

AbdullinAM (changes requested):
Thanks for the implementation! There are some technical issues, nothing too major though. But I would suggest you fix @zarechenskiy's comments first

P.s. If at any point you need help or want to pass over the PR --- I can take over and finalise it myself.

qwwdfsad:
While we are actively working on the new format, I'd suggest postponing this one to minimize the number of disruptive changes for ktfmt users

Issue #640, opened 2026-07-08
Ktfmt 0.64 runs are not idempotent


lwasyl:
The following test fails:

```kotlin
@Test
fun foo() {
  val code =
      """
      |//////////////////////////////////////////
      |fun test() {
      |  scope {
      |    val foo = 
      |      runCatching { someFunction() }.getOrNull()
      |  }
      |}
      |"""
          .trimMargin()

  val options = FormattingOptions(
    preserveLambdaBreaks = true,
    blockIndent = 2,
    continuationIndent = 2,
    maxWidth = code.lineSequence().first().length,
  )
  val formattedOnce = Formatter.format(
    options = options,
    code = code,
  )

  assertThatFormatting(formattedOnce)
      .withOptions(options)
      .isEqualTo(formattedOnce)
}
```

The first formatting produces
```kotlin
fun test() {
  scope {
    val foo = runCatching {
      someFunction()
    }.getOrNull()
  }
}
```

the second one does
```kotlin
fun test() {
  scope {
    val foo = runCatching {
      someFunction()
    }
      .getOrNull()
  }
}
```

(btw kind of surprised that the formatting isn't 
```
val foo = runCatching { someFunction() }
  .getOrNull()
``` 
but I think that's intended? I mean the line break in lambda)

lwasyl:
Fixed in https://github.com/Kotlin/ktfmt/pull/649

PR #649, opened 2026-07-29
Fix non-idempotent runs when lambda with chained call is broken (#640)

lwasyl:
Took a stab at #640, I admit I don't fully grok the AST machinery and had some AI help me with this. But the fix look plausible and all the tests pass


hick209 (approved):
LGTM, but I can no longer run repo-wide tests on PRs as I could before it moved to Kotlin Foundations, so I'll trust them to merge the PR in case things are good and don't regress anything

PR #652, opened 2026-07-31
Upgrade tests to JUnit 6

qwwdfsad:
A few notes:

* kotlin-test is removed as obsolete
* JUnit4 in a lambda module is kept, I'll remove the whole module later, no need to touch it
* @AbdullinAM for your experiments with CI -- the min JDK baseline is now 17 due to JUnit 6 requirement

This is a basis for the next PR with tests revamp (and it's more pleasant to do it with JUnit 6 + it's always nice to renovate things) and a nice testing ground for stacked PRs


core/build.gradle.kts
```diff
@@ -41,9 +41,12 @@ dependencies {
   api(libs.kotlin.stdlib)
   api(libs.kotlin.compilerEmbeddable)
   implementation(libs.ec4j)
-  testImplementation(libs.kotlin.test.junit4)
-  testImplementation(libs.googleTruth)
-  testImplementation(libs.junit)
+  testImplementation(platform(libs.junit.bom))
+  testImplementation(libs.googleTruth) {
+    exclude(group = "junit", module = "junit")
```
qwwdfsad: See https://github.com/google/truth/issues/333 and https://github.com/google/truth/blob/18fafefc293fd624a5291cda8da42356401650eb/core/src/main/java/com/google/common/truth/Platform.java#L264-L280

My suggestion is to get rid of Truth to moment we have any JUnit4/linkage errors in the build

gradle/libs.versions.toml
```diff
@@ -27,11 +27,12 @@ googleJavaformat = { module = "com.google.googlejavaformat:google-java-format",
 googleTruth = { module = "com.google.truth:truth", version.ref = "com-google-truth-truth" }
 gson = { module = "com.google.code.gson:gson", version.ref = "com-google-code-gson-gson" }
 guava = { module = "com.google.guava:guava", version.ref = "com-google-guava-guava" }
-junit = { module = "junit:junit", version.ref = "junit-junit" }
+junit-bom = { module = "org.junit:junit-bom", version.ref = "org-junit-junit-bom" }
```
qwwdfsad, line 20: If you are puzzled by the convention between dashes and camel case, it makes two of us now: https://docs.gradle.org/current/userguide/best_practices_dependencies.html#name_version_catalog_entries

core/build.gradle.kts
AbdullinAM: My suggestion is to just get rid of it. As far as I can see, there is nothing specific about its usage. All tests can use default JUnit assertions

AbdullinAM:
My only concern with this approach to test generation is that it is very inconvenient for debugging individual test cases: you have to add conditional filtering in the factory to run a single test that you are interested in. My original idea was to replicate the approach from the compiler, I even had a [prototype](https://github.com/AbdullinAM/ktfmt/pull/1) that does it. Maybe we can adapt something like this?

qwwdfsad:
@AbdullinAM do you mean from Gradle and/or -`agent:` or from IntelliJ? If the latter, IJ is perfectly capable of selecting one test without any issues, there on rerun

core/build.gradle.kts
qwwdfsad: Done

PR #655, opened 2026-08-03
Introduce FormattingOptions.useExperimentalEngine

qwwdfsad:
* Private, internal-only mode for ongoing development
* A separate KotlinInputAstVisitor to evolve independently, so no actual users are affected
* Add a single test case to showcase the idea

The idea is that we can tweak the independent, publicly non-accessible visitor/format to see if we can comply with various findings from our [comparison](https://htmlpreview.github.io/?https://github.com/qwwdfsad/kotlin-format/blob/master/report.html) by adjusting rules here and there


core/src/main/java/com/facebook/ktfmt/format/KotlinInputAstVisitor.kt
```diff
@@ -2144,7 +2146,9 @@ class KotlinInputAstVisitor(
         visit(psi)
       }
 
-      if (onlyAnnotationsSoFar) {
+      if (onlyAnnotationsSoFar && forceAnnotationBreaks && psi is KtAnnotationEntry) {
```
qwwdfsad, line 2149: Maybe it's worth copying and pasting the entire method to avoid _any_ inference. 
A bit harder to maintain, but not much difference

PR #656, opened 2026-08-04
Remove kotlin.github.io/ktfmt

qwwdfsad:
Rationale:

* It relies on Meta-hosted AWS, which will eventually be shut down
* It has "Facebook OSS" and a copyright identity. It feels wrong to just replace it, it is wrong to operate under it from the outside
* It's not in the search output (I've checked DuckDuckGo and Google Search)

Eventually we can get back to that with Dokka/md and, instead of lambda, either with GraalVM+Wasm or with K/Wasm if we decide to migrate from PSI to KMP parser


qwwdfsad:
cc @hick209

AbdullinAM (approved):

PR #657, opened 2026-08-04
TeamCity CI fixes

AbdullinAM:
Motivation:
* Gradle 8.14 is only compatible with JDKs up to 24, quite a lot of people already use JDK 25+ by default, including our TC agents
* Gradle 9.6.1 is compatible with JDKs 17-26, which matches our build requirements
* Native image plugin change is necessary for Gradle 9.0+ compatibility
* Test changes remove OS-specific elements for compatibility with Windows agents


qwwdfsad (approved):
If our CI covers GitHub actions, maybe it's time to disable the latter so the build status will be less misleading

AbdullinAM:
Addressed GH Actions in #658

PR #663, opened 2026-08-10
Configure native image CI

AbdullinAM:


settings.gradle.kts
```diff
@@ -44,3 +44,7 @@ dependencyResolutionManagement {
     mavenCentral()
   }
 }
+
+plugins {
+  id("org.gradle.toolchains.foojay-resolver-convention") version "1.0.0"
```
AbdullinAM, line 49: 1. GraalVM image on CI does not have preinstalled JDK 17, this is a temporary workaround to make it work while we set up our own docker images
2. Nice to have for anyone who does not have JDK 17 toolchain configured

qwwdfsad (approved):

PR #664, opened 2026-08-10
Migrate packages

AbdullinAM:


CHANGELOG.md
```diff
@@ -15,178 +15,178 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
   selected-range formatting flags.
 - Support selected-range formatting in the IntelliJ plugin.
 - GraalVM native image support
-- Support val/var keywords in destructuring entries (https://github.com/facebook/ktfmt/pull/637)
+- Support val/var keywords in destructuring entries (https://github.com/Kotlin/ktfmt/pull/637)
```
qwwdfsad: Worth keeping the changelog for already happened releases as it was

qwwdfsad (changes requested):

qwwdfsad:
While we are here, could you please also rename `java` -> `kotlin` source folder?

CHANGELOG.md
AbdullinAM: Ah, the old links are still redirected correctly, so yeah makes sense to keep them

qwwdfsad (approved):

PR #665, opened 2026-08-10
[build] Replace `native_smoke_test.sh` script with a Gradle task

AbdullinAM:


AbdullinAM:
I updated the CI to work with the new Gradle task (its easily revertable if we don't go ahead with this change)

qwwdfsad (approved):

PR #667, opened 2026-08-11
Migrate formatting tests to file-based approach

AbdullinAM:
* Add support for per-case directives
* Migrate formatting tests to file-based approach

Note: The total amount of tests has changed from 797 to 867. This difference comes from:
1. **New idempotency check tests**. For each file-based test that has a separate `.output` file we also generate an additional `Format is idempotent` test. This accounts for 100 new tests
2. **Merged/removed tests**. During the manual analysis and migration of test cases some of them were merged into a single file (when they test the same concept) or removed (if the test duplicates already existing one). This accounts for -30 total tests

Note 2: `KDocFormatterTest` is still left in its original form, it will require more handling and will be addressed in a separate commit later. This PR is just migration of tests that I have already processed in a separate fork


core/src/main/kotlin/org/jetbrains/ktfmt/format/FormattingOptions.kt
```diff
@@ -216,6 +218,11 @@ data class FormattingOptions(
           this.debuggingPrintOpsAfterFormatting = debuggingPrintOpsAfterFormatting
         }
 
+    /** @see FormattingOptions.experimentalEngine */
```
qwwdfsad: Definitely redundant comment

qwwdfsad:
Note: we'll check if `.git-blame-ignore-revs` works for this type of changes

core/src/test/kotlin/org/jetbrains/ktfmt/testutil/CaseConfig.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Parses the leading comment block of [code] as a directive header. [origin] can be added for
+     * better error messages
+     */
+    fun parse(code: String, origin: Path? = null): ParsedDirectives {
+      val header = code.lines().takeWhile { it.startsWith("//") }
+      val parsedDirectives =
+          header
+              .mapIndexedNotNull { lineNumber, line ->
```
qwwdfsad, line 48: I would expect test failure on an incorrect directive rather than skipping it and making it part of the test

qwwdfsad (commented):
I haven't manually looked through every test case obviously, only infra, I assume you've used some kind of deterministic automatic migration for that (worth noting in the squashed commit message for the history sake).

I see there are around ~70 new tests (863 vs 797), could you please elaborate on where they came from?

core/src/test/kotlin/org/jetbrains/ktfmt/testutil/CaseConfig.kt
AbdullinAM: In that case we would not allow to have any comments in the header which will break several test cases like [newLineBetweenComments2.input](https://github.com/Kotlin/ktfmt/blob/abdullin/tests/core/src/test/resources/cases/format/comment/newLineBetweenComments2.input). Although I agree that this setup is a bit janky right now, I will think how to improve it.

AbdullinAM: Implemented something in between: we fail with error if we find a directive after a regular comment in the header

qwwdfsad (approved):
+7000, -11000 LGTM 👍 

(that's me trying to be funny, all good)

PR #668, opened 2026-08-11
Visitor refactoring

AbdullinAM:
Motivation: simplify `KotlinInputAstVisitor` by separating it into separate independent formatter visitors responsible for small subsystems. This is an initial prototype, the separation will continue in the following PRs

Note: this is just general idea for the structure, I haven't refactored the 'contents' of the moved functions. I plan to do that after the split


core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/AbstractFormatterVisitor.kt
```diff
@@ -0,0 +1,416 @@
...
+import org.jetbrains.kotlin.psi.KtUserType
+import org.jetbrains.kotlin.psi.KtValueArgument
+import org.jetbrains.kotlin.psi.KtValueArgumentList
+import org.jetbrains.kotlin.psi.KtWhenConditionInRange
+import org.jetbrains.kotlin.psi.KtWhenConditionIsPattern
+import org.jetbrains.kotlin.psi.KtWhenConditionWithExpression
+import org.jetbrains.kotlin.psi.KtWhenExpression
+import org.jetbrains.kotlin.psi.KtWhileExpression
+
+abstract class AbstractFormatterVisitor : KtTreeVisitorVoid(), KotlinAstFormatter {
```
AbdullinAM, line 85: This adds an overhead to the AST visitor, but my intuition is that right now it makes sense. I did not notice any measurable degradation on coroutines

qwwdfsad: (I haven't started the review yet) not that you have to comment out `parallelStream()` in `Main` -- it's the main source of non-determinism for local measurements

AbdullinAM: Manual experiments on `kotlinx.coroutines` and my local JMH benchmarks (on real and synthetic samples) did not show any measurable difference.

This confirms my intuition: with this one more layer of visitor we basically add one more function call in the visitor call chain, that does not have a large effect on the final performance

AbdullinAM (commented):
@qwwdfsad Your changes look good to me 👍

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/KotlinAstFormatter.kt
```diff
@@ -0,0 +1,394 @@
...
+
+  fun formatTypeParameter(parameter: KtTypeParameter)
+
+  fun formatTypeConstraintList(list: KtTypeConstraintList)
+
+  fun formatTypeConstraint(constraint: KtTypeConstraint)
+
+  fun formatFunctionType(type: KtFunctionType)
+
+  fun formatClassOrObject(classOrObject: KtClassOrObject) {}
```
qwwdfsad: One thing that bothers me -- by looking at the code, I cannot really quickly determine whether the empty implementation will actually delete the code or is unused.

Maybe let's do something like `TODO("Unreachable code path")` as the deafult impl?

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/KotlinAstFormatter.kt
```diff
@@ -0,0 +1,394 @@
...
+
+  fun formatDelegatedSuperTypeEntry(specifier: KtDelegatedSuperTypeEntry) {}
+
+  fun formatClassBody(body: KtClassBody) {}
+
+  fun formatValueArgumentList(list: KtValueArgumentList): BreakTag?
+
+  fun formatModifierList(list: KtModifierList)
+
+  fun formatContextReceiverList(contextReceiverList: KtContextReceiverList)
```
qwwdfsad, line 214: Here (and around `visitContextReceiverList`), every caller but one now kind of has to remember to insert a linebreak after the call.

Maybe the function itself should own that?

qwwdfsad: On a side note, I have no idea what is going to happen if we decide to use the line ranges API around this part. 

My bet is that it's going to produce incorrect formatting because the break after `context()` is kind of atomic with the modifier itself, while `range` operates on non-atomic units.
Which means that either formatting ops should operate on atomic ranges or we should reconsider the range flag itself

AbdullinAM: Yes, I already have a version doing that. My intuition with this was: lets introduce the general structure in this PR, and change the behaviour/refactor the code in separate PRs. Otherwise it will be too hard to track changes IMO

qwwdfsad: Sure

qwwdfsad (approved):

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/KotlinAstFormatter.kt
AbdullinAM: > what is going to happen if we decide to use the line ranges API around this part

I think we will need to extend our test suite to see, because right now it is virtually not tested at all. Considering that it is the main entry point for IDE

hick209:
And so it began...

Awesome work here, I'm excited to see the new format! 😃

PR #670, opened 2026-08-14
Support editorconfig when using stdin

ZacSweers:
This PR adds support for, when formatting source from stdin, use `--stdin-name` as the logical source path for `EditorConfig` lookup. This lets tools format virtual or staged source content while preserving ancestor `EditorConfig` settings and path-specific sections.

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/ktfmt/cli/ParsedArgs.kt
+++ b/core/src/main/kotlin/org/jetbrains/ktfmt/cli/ParsedArgs.kt
@@ -37,7 +37,7 @@ data class ParsedArgs(
 
     /** Return exit code 1 if any formatting changes are detected. */
     val setExitIfChanged: Boolean,
-    /** File name to report when formating code from stdin */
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/org/jetbrains/ktfmt/cli/ParsedArgs.kt
```diff
@@ -88,7 +88,8 @@ data class ParsedArgs(
         |  --meta-style                      Use 2-space block indenting (default)
         |  --google-style                    Google internal style (2 spaces)
         |  --kotlinlang-style                Kotlin language guidelines style (4 spaces)
-        |  --stdin-name=<name>               Name to report when formatting code from stdin
+        |  --stdin-name=<path>               Path to report for stdin and, when EditorConfig is
```
AbdullinAM, line 91: I would suggest keeping the original help message and just add a second part to it: `Path to report when formatting code from stdin, also used for EditorConfig lookup if enabled`. Just `Path to report for stdin` is a bit vague

AbdullinAM (approved):

PR #673, opened 2026-08-14
Small cleanup

AbdullinAM:
Some small cleanups that are not entirely connected with each other rather that just being general refactorings and simplifications done while analyzing the code

* Extract more utility methods from KIAV to PsuUtils
* Simplify some of the TypeFormatter methods
 * delegate TypeFormatter::formatFunctionType.parameterList formatting to dedicated method
* Use `formatCommaSeparatedList` in more cases where it can be used as a drop-in replacement. Other cases (e.g. `visitFunctionLikeExpression`) require more structural changes
* Extract a separate `formatInitializerExpression` method for formatting everything that comes after `=`

### DIFF
--- a/core/api/ktfmt.api
+++ b/core/api/ktfmt.api
@@ -183,8 +183,11 @@ public final class org/jetbrains/ktfmt/format/KotlinInput : com/google/googlejav
 	public fun getkN ()I
 }
 
-public class org/jetbrains/ktfmt/format/KotlinInputAstVisitor : org/jetbrains/ktfmt/format/visitor/AbstractFormatterVisitor, org/jetbrains/ktfmt/format/visitor/FileFormatter, org/jetbrains/ktfmt/format/visitor/ListFormatter, org/jetbrains/ktfmt/format/visitor/TypeFormatter {
+public class org/jetbrains/ktfmt/format/KotlinInputAstVisitor : org/jetbrains/ktfmt/format/visitor/AbstractFormatterVisitor, org/jetbrains/ktfmt/format/visitor/ExpressionFormatter, org/jetbrains/ktfmt/format/visitor/FileFormatter, org/jetbrains/ktfmt/format/visitor/ListFormatter, org/jetbrains/ktfmt/format/visitor/TypeFormatter {
 	public fun <init> (Lorg/jetbrains/ktfmt/format/FormattingOptions;Lcom/google/googlejavaformat/OpsBuilder;)V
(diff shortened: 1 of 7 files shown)


hick209 (dismissed):
LTGM, assuming this is just house ordering, which is what it looked like to me, but since it's a big change I know I might have missed something.

core/api/ktfmt.api
```diff
@@ -562,11 +562,17 @@ public final class org/jetbrains/ktfmt/format/visitor/OpsUtilsKt {
 public final class org/jetbrains/ktfmt/format/visitor/PsiUtilsKt {
 	public static final fun getCallExpression (Lorg/jetbrains/kotlin/psi/KtExpression;)Lorg/jetbrains/kotlin/psi/KtCallExpression;
 	public static final fun getChainRoot (Lorg/jetbrains/kotlin/psi/KtExpression;)Lorg/jetbrains/kotlin/psi/KtExpression;
-	public static final fun getHasEmptyParens (Lorg/jetbrains/kotlin/psi/KtValueArgumentList;)Z
-	public static final fun hasEmptyParens (Lorg/jetbrains/kotlin/psi/KtParameterList;)Z
+	public static final fun getHasEmptyParenthesis (Lorg/jetbrains/kotlin/psi/KtParameterList;)Z
```
qwwdfsad, line 579: At this point, it's just a nuisance.
Let's handle this separately and exclude everything but what actually consitutes ABI

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/PsiUtils.kt
```diff
@@ -110,6 +114,26 @@ val KtExpression.chainRoot: KtExpression
     return root
   }
 
+/**
+ * Decomposes a qualified expression into parts, so `rainbow.red.orange.yellow` becomes `[rainbow,
+ * rainbow.red, rainbow.red.orange, rainbow.orange.yellow]`
+ */
+val KtExpression.chainParts: List<KtExpression>
```
qwwdfsad, line 121: Is this one unused?

qwwdfsad (dismissed):

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/PsiUtils.kt
AbdullinAM: Yes, moved the util itself and forgot to replace its usage. Will fix

qwwdfsad (approved):

PR #674, opened 2026-08-14
Extract AnnotationFormatter and update KotlinLang annotation formatting style

AbdullinAM:
* Update KotlinLang style annotations formatting rules for declarations and types
* Keep the formatting of annotations on expressions the same as default (no forced breaks), except one thing: annotations are blocked together with the expression itself, so if the formatting of an expression does not fit into a single line, each annotation is forced to a new line too. See last example from `AnnotationOnExpression.input`

Motivation: 
In my opinion current formatting of annotations on expressions is pretty idiomatic. There are some one-off examples where I would prefer to have a different formatting (e.g. first case from `AnnotationOnExpression.input`), but I can't come up with a consistent rule that will distinguish this case from any other. If you have an opinion or want more examples --- I can provide them

Note: 
I think it will also make sense to just test new formatter style on all existing tests as well, that way we will be able to see how new style affects things easier


hick209 (dismissed):

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangAnnotationFormatter.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.jetbrains.kotlin.psi.KtBinaryExpression
+import org.jetbrains.kotlin.psi.KtBinaryExpressionWithTypeRHS
+import org.jetbrains.kotlin.psi.KtBlockExpression
+import org.jetbrains.kotlin.psi.KtLambdaExpression
+import org.jetbrains.kotlin.psi.KtReturnExpression
+import org.jetbrains.ktfmt.format.visitor.AnnotationFormatter
+import org.jetbrains.ktfmt.format.visitor.block
+import org.jetbrains.ktfmt.format.visitor.sync
+
+interface KotlinLangAnnotationFormatter : AnnotationFormatter {
```
qwwdfsad, line 34: Taking, we are going to decompose/read, and later maintain KL*F quite a lot, and it's going to diverge from the baseline, let's maybe have some convention for the documentation?

We can, of course, deduce the behaviour from implementation/tests, but it would be nice to have something in the class/method KDoc that immediately explains what we can expect from the output. Shouldn't be medically-precise (there are like 25 places where you can put an annotation) but definitely should give the rought idea

core/src/test/resources/cases/new_codestyle/annotation/AnnotationsOnParameters.output
```diff
@@ -0,0 +1,28 @@
...
+  fun wrapped(
+      @NotNull @Size(min = 1, max = 100) someVeryLongParameterName: String,
+      @Nullable anotherRatherLongParameterName: String? = null,
+  ) = someVeryLongParameterName + anotherRatherLongParameterName
+
+  fun varargs(@NotNull vararg values: String) = values.size
+
+  fun higherOrder(@NotNull block: (@Nullable Int) -> Unit) = block(1)
+
+  fun @receiver:NotNull String.extension(@NotNull other: String) = this + other
```
qwwdfsad: TIL you can do that

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangListFormatter.kt
```diff
@@ -38,7 +40,14 @@ interface KotlinLangListFormatter : ListFormatter {
         format(psi)
       }
 
-      if (onlyAnnotationsSoFar && psi is KtAnnotationEntry) {
+      val shouldForceBreak =
+          // don't force break on parameter annotations
+          list.parent !is KtParameter &&
```
qwwdfsad: I feel very uneasy about this check: it's hard to reason about, it's hard to ensure nothing is missed, and that it's evolved properly in the presence of various syntactic combinations.

E.g. with a new playground, the following breaks [1].
I _assume_ it's because this check is incomplete

Image

qwwdfsad:
```
  
val foo: @Ann String = ""  
  
val foo: @Ann("") String = ""  
  
val foo: @Ann String? = null  
  
val foo: List<@Ann Int> = listOf()  
  
val veryLongType: @Ann("with an argument") Map<String, List<Int>> = emptyMap()  
```

text for you to copy-paste

qwwdfsad (changes requested):

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangListFormatter.kt
qwwdfsad: The comment here (see above about the convention) with a proper semantic meaning probably would've helped to spot it earlier

AbdullinAM: I reworked the formatting rules, now they list all cases that **do** force a break, instead of listing all exceptions. Hopefully this (with updated documentation) will make it cleaner

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangAnnotationFormatter.kt
```diff
@@ -3,15 +3,34 @@ package org.jetbrains.ktfmt.format.visitor.kotlinlang
...
 import org.jetbrains.ktfmt.format.visitor.AnnotationFormatter
 import org.jetbrains.ktfmt.format.visitor.block
+import org.jetbrains.ktfmt.format.visitor.isBinaryExpression
 import org.jetbrains.ktfmt.format.visitor.sync
 
+/**
+ * Custom annotation formatter for KotlinLang style. Currently only modifies formatting of
+ * annotations on expressions via [formatAnnotatedExpression].
+ *
+ * General rule for annotations on an expression: keep annotations on the same line as the
```
qwwdfsad, line 18: Fun fact: not everything that's on RHS is an expression.

It means that for
```
val x = @Foo fun (a: Int, b: Int) = a + b
```

we will force a line break.

Not actionable, rather sharing a pretty weird use-case

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangAnnotationFormatter.kt
```diff
@@ -3,15 +3,34 @@ package org.jetbrains.ktfmt.format.visitor.kotlinlang
...
 import org.jetbrains.kotlin.psi.KtBlockExpression
 import org.jetbrains.kotlin.psi.KtLambdaExpression
 import org.jetbrains.kotlin.psi.KtReturnExpression
 import org.jetbrains.ktfmt.format.visitor.AnnotationFormatter
 import org.jetbrains.ktfmt.format.visitor.block
+import org.jetbrains.ktfmt.format.visitor.isBinaryExpression
 import org.jetbrains.ktfmt.format.visitor.sync
 
+/**
+ * Custom annotation formatter for KotlinLang style. Currently only modifies formatting of
```
qwwdfsad: (here and for the PR description) Let's be a bit more explicit about the goals:

E.g. 

* " Custom annotation formatter for KotlinLang style" -> "Custom expression annotation formatter for KotlinLang style. For statements, declarations and types see ..."
* PR/commit descriptions here

qwwdfsad: Mostly need that to be explicit about whether things like
```
@Serializable(with = RegexSerializer::class) kotlin.text.Regex
...
@Foo
val a = 2
```

in the exploration scope

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangAnnotationFormatter.kt
AbdullinAM: TBH I'm fine with having a line break here, it keeps it more consistent with the overall rules. But if we want to change that --- I can update the code

qwwdfsad (dismissed):
Looks good to me, thanks!

I'll ask Misha or Pasha to review it as well and then let's merge it

core/src/test/resources/cases/new_codestyle/annotation/AnnotationOnExpression.new.output
```diff
@@ -0,0 +1,90 @@
...
+        @Suppress("WHEN")
+        when (value) {
+            else ->
+                compute(
+                    alpha,
+                    beta,
+                    gamma,
+                    delta,
+                    epsilon,
+                )
```
zarechenskiy: Minor but maybe we can increase MAX_WIDTH here to make sure that the code doesn't affect arguments as in the "old" version we have just `compute(alpha, beta, gamma...`

zarechenskiy (dismissed):

core/src/test/resources/cases/new_codestyle/annotation/AnnotationOnExpression.new.output
AbdullinAM: Increasing `MAX_WIDTH` causes too many changes in the other parts of the test. I removed the last argument of the `compute` call, now they're similar

core/src/test/resources/cases/new_codestyle/annotation/AnnotationOnExpression.new.output
```diff
@@ -0,0 +1,83 @@
+// MAX_WIDTH 55
+
+private fun Window.animationQueue(): AnimationQueue =
+    @Suppress("UnsafeCastFromDynamic")
```
AbdullinAM, line 4: This code will be formatted like 
```
private fun Window.animationQueue(): AnimationQueue =
    @Suppress("UnsafeCastFromDynamic") asDynamic().queue
        ?: AnimationQueue(this).also {
            asDynamic().queue = it
        }
```

if we increase the max width. Because this is a binary expression `?:` and we don't force breaks in binary expressions

zarechenskiy: Let's increase the width then to see further potential changes?

zarechenskiy (dismissed):

zarechenskiy (approved):

PR #675, opened 2026-08-17
ktfmt offline playground

qwwdfsad:
`./playground/run.sh`

Something like this is very convenient for reviewing code style adjustments.
I've tried to review the actual _codestyle_ change in #674, but it's really tough without having an interactive playground.

I originally tried to build it with Wasm, but for K/W we have to fully migrate to the KMP parser, and for NI it doesn't work for a variety of reasons (e.g., https://github.com/oracle/graal/issues/14193). So maybe next time :)


AbdullinAM (approved):
Really nice quality of life feature! Thanks!

Notes for the future improvement: I think it will also fun to rebuild this with full Kotlin/Gradle and Compose Web and also test the speed of pure JVM/in-process formatting: maybe the benefit of the native image speed is not always worth extra 30 second wait on startup.

PR #676, opened 2026-08-17
Add --range-start and --range-end CLI options for range formatting

### LINKED ISSUE #573: Reformat before commit and partial reformat is not supported from IntelliJ plugin

samccone:
Implement character offset-based range formatting inspired by Prettier's range options (--range-start and --range-end).

Features:
- `--range-start <int>`: start offset (inclusive, defaults to 0)
- `--range-end <int>`: end offset (exclusive, defaults to EOF)
- Support independent or paired usage with robust bounds validation
- Safe range clamping in Formatter and KotlinInput to prevent out-of-bounds errors
- Shebang offset adjustment and line-ending preservation
- Expanded unit and integration test coverage across CLI, Tokenizer, and Formatter

Fixes: #573, #388

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/ktfmt/cli/ParsedArgs.kt
+++ b/core/src/main/kotlin/org/jetbrains/ktfmt/cli/ParsedArgs.kt
@@ -47,6 +47,7 @@ data class ParsedArgs(
   internal val lineRanges: RangeSet<Int> = TreeRangeSet.create()
   /** Zero-indexed character ranges to format, using closed-open bounds. */
   internal val characterRanges: RangeSet<Int> = TreeRangeSet.create()
+  internal var isPartialFormatting: Boolean = false
 
   companion object {
 
@@ -96,6 +97,10 @@ data class ParsedArgs(
         |  --length=<length>                 Character length to format, paired with --offset.
         |                                        May be used multiple times. 0 formats the whole
         |                                        line containing the given --offset.
+        |  --range-start=<offset>            Format code starting at a given character offset
+        |                                        (inclusive). Defaults to 0.
+        |  --range-end=<offset>              Format code ending at a given character offset
+        |                                        (exclusive). Defaults to end of file.
         |  --set-exit-if-changed             Sets exit code to 1 if any input file was not
         |                                        formatted/touched
         |  --do-not-remove-unused-imports    Leaves all imports in place, even if not used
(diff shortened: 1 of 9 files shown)


core/src/main/kotlin/org/jetbrains/ktfmt/cli/ParsedArgs.kt
```diff
@@ -96,6 +97,10 @@ data class ParsedArgs(
         |  --length=<length>                 Character length to format, paired with --offset.
         |                                        May be used multiple times. 0 formats the whole
         |                                        line containing the given --offset.
+        |  --range-start=<offset>            Format code starting at a given character offset
```
AbdullinAM: The same functionality is already supported by ktfmt via `--offset` and `--length` options. What is your motivation/use case for adding new CLI options?

AbdullinAM (changes requested):
Hi! Thank you for the contribution! I have some comments on the PR, but before doing a full review, can you please explain your motivation/use case for adding new CLI options for functionality that is already supported by ktfmt via `--offset` and `--length`?

PR #678, opened 2026-08-19
Revert formatting of block-like calls

AbdullinAM:
Extract the behaviour introduced in #634 into KotlinLang style formatters and revert the default formatter. Motivation: we don't want to change the behaviour of the existing formatter while we're also evolving the new Kotlin Lang style.


qwwdfsad (dismissed):
Thanks!

Please conside reopening the original issue as well.

PR #683, opened 2026-08-26
Use file extensions to distinguish between regular .kt and script .kts files

AbdullinAM:
Motivation: previously `Parser.parse` parsed everything like a Kotlin script (kts) file. This results in the source code files being represented slightly differently, e.g.:
* Context receivers on top-level declarations are not parsed properly
* Top-level properties are actually represented as local properties
* etc.

While this does not introduce any immediate downsides (except, arguably, for the context receivers case) and the formatter works, it produces unexpected parsing results and may cause confusion in the future. To prevent that, we actually resolve the file's extension and use it to correctly choose the parsing mode.

Changes:
* Introduce new `KotlinCode` class that represents a Kotlin source code file
* Update Formatter and Parser APIs to accept `KotlinCode` instances and/or string code representation with explicit file type (for tests)
* Update test infrastructure to support new `// FILE_TYPE` directive, update tests to use `REGULAR` file type by default
* Simplify Formatter API around line/charachter ranges:
  * Remove unnecessary manual handling of shebang lines
  * Formatter only accept charachter ranges, line ranges are converted to char ranges on the call site through `KotlinCode.lineRangesToCharRanges`
* Unify `FileFormatter` methods to handle `KtFile` and `KtScript` similarly


core/src/test/resources/cases/format/call/nestedCallsKotlinlangStyle.input
```diff
@@ -1,7 +1,6 @@
 // MAX_WIDTH 70
-// BLOCK_INDENT 4
-// CONTINUATION_INDENT 4
```
kunyavskiy, line 3: Why did we chnge the test?

core/src/test/resources/cases/format/call/shortCallChainWithLambda2.output
```diff
@@ -1,5 +1,7 @@
 // MAX_WIDTH
 // 13
+// FILE_TYPE
+// SCRIPT
```
kunyavskiy, line 4: That doesn't look good

core/src/test/resources/cases/format/call/nestedCallsKotlinlangStyle.input
AbdullinAM: Originally test was called `nestedCallsKotlinlangStyle`: it manually encoded the kotlinlang formatting options, but was in the same folder as 'default' meta-style tests. I moved it into the `kotlinlang` folder that applies kotlinlang formatting options by default

core/src/test/resources/cases/format/call/shortCallChainWithLambda2.output
AbdullinAM: Missed, will fix. Thanks

core/src/test/kotlin/org/jetbrains/ktfmt/testutil/CaseConfig.kt
```diff
@@ -1,23 +1,28 @@
...
     var checkIdempotency: Boolean = true
+    var fileType = FileType.REGULAR
 
     fun build(): CaseConfig = CaseConfig(options.build(), checkIdempotency)
   }
 
   companion object {
     private val DIRECTIVES: Map<String, Directive> = listOf(
-        Directive("MAX_WIDTH") { options.maxWidth(it.toInt()) },
+        Directive("FILE_TYPE") {} Directive ("MAX_WIDTH") { options.maxWidth(it.toInt()) },
```
kunyavskiy: Why is this even compilable?

core/src/main/kotlin/org/jetbrains/ktfmt/format/Parser.kt
```diff
@@ -60,14 +60,18 @@ object Parser {
...
+    val virtualFile =
+        LightVirtualFile("temp.${code.fileType.extension}", KotlinFileType.INSTANCE, code.code)
     val ktFile = PsiManager.getInstance(env.project).findFile(virtualFile) as KtFile
     val descendants = ktFile.collectDescendantsOfType<PsiErrorElement>()
-    if (descendants.isNotEmpty()) throwParseError(code, descendants[0])
+    if (descendants.isNotEmpty()) throwParseError(code.code, descendants[0])
     return ktFile
   }
 
+  /** Only for tests */
```
kunyavskiy: Probably better to be opt-in than

core/src/main/kotlin/org/jetbrains/ktfmt/util/ranges.kt
```diff
@@ -0,0 +1,11 @@
+package org.jetbrains.ktfmt.util
+
+import com.google.common.collect.RangeSet
+import com.google.common.collect.TreeRangeSet
+
+operator fun <T : Comparable<T>> RangeSet<T>.plus(other: RangeSet<T>): TreeRangeSet<T> {
```
kunyavskiy: Should we actually expose it's TreeRangeSet, not just RangeSet, if arguments were just a RangeSet?

kunyavskiy (changes requested):

core/src/test/resources/cases/format/call/shortCallChainWithLambda2.output
AbdullinAM: Ah, I didn't notice initially. This is the expected output file: initial directive is broken into two lines because of the small max_width

core/src/main/kotlin/org/jetbrains/ktfmt/format/Parser.kt
```diff
@@ -60,14 +61,18 @@ object Parser {
...
+    val virtualFile =
+        LightVirtualFile("temp.${code.fileType.extension}", KotlinFileType.INSTANCE, code.code)
     val ktFile = PsiManager.getInstance(env.project).findFile(virtualFile) as KtFile
     val descendants = ktFile.collectDescendantsOfType<PsiErrorElement>()
-    if (descendants.isNotEmpty()) throwParseError(code, descendants[0])
+    if (descendants.isNotEmpty()) throwParseError(code.code, descendants[0])
     return ktFile
   }
 
+  @InternalKtfmtTestApi
```
AbdullinAM, line 73: This is a very basic and very limited configuration of opt-ins. We will need to extend that, but I think that it is out of the scope of this PR. Will do it in the future

core/src/main/kotlin/org/jetbrains/ktfmt/format/KotlinCode.kt
```diff
@@ -0,0 +1,100 @@
...
+class KotlinCode
+private constructor(
+    val code: String,
+    val fileType: FileType,
+    val lineSeparator: String,
+) {
+  fun copy(code: String) = KotlinCode(code, fileType, lineSeparator)
+
+  override fun toString(): String = buildString {
+    for ((index, line) in code.lineSequence().withIndex()) {
```
kunyavskiy: Nit: isn't it equivalent to code.lineSequence().joinToString(lineSeparator, suffix=lineSeparator)?

kunyavskiy (dismissed):

core/src/main/kotlin/org/jetbrains/ktfmt/format/KotlinCode.kt
AbdullinAM: Yeah, it should be. Makes sense, will update

PR #687, opened 2026-08-31
Add a KotlinLang formatter for when expressions

AbdullinAM:
* Don't force break after comma in multiple-condition entries
* Don't allow break before `->` in any cases

### DIFF
--- a/core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangControlFlowExpressionFormatter.kt
+++ b/core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangControlFlowExpressionFormatter.kt
@@ -64,10 +64,12 @@ internal class KotlinLangControlFlowExpressionFormatterImpl : ControlFlowExpress
               builder.token("else")
             } else {
               val conditions = whenEntry.conditions
+              val lastIndex = conditions.lastIndex
+              val lastCommaIndex = if (forceMultiline) lastIndex + 1 else lastIndex
               for ((index, condition) in conditions.withIndex()) {
                 format(condition)
-                builder.guessToken(",")
-                if (index != conditions.lastIndex) {
+                if (index < lastCommaIndex) builder.token(",")
+                if (index < lastIndex) {
                   if (forceMultiline) {
                     builder.forcedBreak()
                   } else {
--- a/core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangControlFlowExpressionFormatter.kt
+++ b/core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangControlFlowExpressionFormatter.kt
@@ -53,6 +53,7 @@ internal class KotlinLangControlFlowExpressionFormatterImpl : ControlFlowExpress
 
       expression.entries.forEachIndexed { index, whenEntry ->
         builder.block(blockIndent) {
+          val forceMultiline = whenEntry.trailingComma != null
           if (index != 0) {
             // preserve new line if there's one
             builder.blankLineWanted(OpsBuilder.BlankLineWanted.PRESERVE)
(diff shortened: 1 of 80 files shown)


core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangControlFlowExpressionFormatter.kt
```diff
@@ -0,0 +1,102 @@
...
+ * when (x) {
+ *     1, 2 -> false
+ *     is LocalVariableDescriptor,
+ *     is ValueParameterDescriptor,
+ *     is ReceiverParameterDescriptor -> true
+ * }
+ * ```
+ * - Never add a break before `->`, even when the condition ends with a trailing comma
+ */
+internal class KotlinLangControlFlowExpressionFormatterImpl : ControlFlowExpressionFormatterImpl() {
```
kunyavskiy, line 45: I don't think it's really important right now, more for todo-list. 

When you have a class with lots of open methods, and overriding some of them, it's becoming hard to keep track on what's actually called when. 

I would really suggest creating a CommonControlFlowExpressionFormatterImpl and inheriting 2 separate classes from them for diffrent styles, instead of inheriting diffrent styles one from another..

core/src/test/resources/cases/new_codestyle/when/guardComments.new.output
```diff
@@ -0,0 +1,9 @@
+fun f(x: Any) {
+    when (x) {
+        is Foo if // line comment after if
+        x.isReady -> doSomething()
```
kunyavskiy: We definetely need identation here.

core/src/test/resources/cases/new_codestyle/when/trailingCommaPreserved.new.output
```diff
@@ -0,0 +1,11 @@
+// MAX_WIDTH 50
+// TRAILING_COMMA_STRATEGY NONE
+
+fun f(x: Any) {
+    when (x) {
+        0, 1, -> print(1)
```
kunyavskiy: Tbh, with trailing comma, it makes sense to write result on a separate line, not doing this kinda kills the reason to use trailing comma - not having extra changes, when you add something here. What do you think about following rule:

* If no trailing comma is present - never break before ->
* If there is trailing comma -> break after trailing comma, have an extra identaion layer before ->

The only thing I'm unsure about is how forcing traling commas mode should work. Maybe it should be a separate setting relative from trailing commas in calls.

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangControlFlowExpressionFormatter.kt
```diff
@@ -0,0 +1,102 @@
...
+          }
+          builder.forcedBreak()
+          builder.block {
+            if (whenEntry.elseKeyword != null) {
+              builder.token("else")
+            } else {
+              val conditions = whenEntry.conditions
+              for ((index, condition) in conditions.withIndex()) {
+                format(condition)
+                builder.guessToken(",")
```
kunyavskiy: Isn't it making comma optional between the two conditions?

AbdullinAM: I don't think it makes it optional, but I agree that it should be rewritten to make it more obvious

core/src/test/resources/cases/new_codestyle/when/guardComments.new.output
AbdullinAM: Yes, we do. But this problem is not necessarily connected with the when formatting, so I would prefer to handle it in a separate PR. I created an issue for it #692

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangControlFlowExpressionFormatter.kt
```diff
@@ -64,10 +64,12 @@ internal class KotlinLangControlFlowExpressionFormatterImpl : ControlFlowExpress
               builder.token("else")
             } else {
               val conditions = whenEntry.conditions
+              val lastIndex = conditions.lastIndex
+              val lastCommaIndex = if (forceMultiline) lastIndex + 1 else lastIndex
```
kunyavskiy, line 68: Minor: probably checking on trailing lambda explicitly is better, in case we would have other reasons to make multiline later.

kunyavskiy (dismissed):

PR #691, opened 2026-09-03
New formatting style for assignment-like expressions

AbdullinAM:
This PR implements a new rule for formatting assignment-like expressions in KotlinLang style. General idea: preserve user-provided line break after the assignment operator.

The most complicated part of the PR is the handling of chained calls, especially when they are preceded by a lambda:

```kotlin
val foo: Int
    get() = run {
        bar()
        baz()
    }
        .foo()
        .bar()
```

We considered several alternatives, main ones are:
* Don't add indent for the chained calls:
```kotlin
val foo: Int
    get() = run {
        bar()
        baz()
    }
    .foo()
    .bar()
```

* Don't force break after the closing `}`:
```kotlin
val foo: Int
    get() = run {
        bar()
        baz()
    }.foo()
        .bar()
```

* Add the indent for the whole chain, including the body of the first call:
```kotlin
val foo: Int
    get() = run {
            bar()
            baz()
        }
        .foo()
        .bar()
```

I decided to implement the latter one because it behaves most consistently in all cases that I was able to come up with. It is also very similar to the current default style.

#### Known issues

* Annotated expressions don't have a continuation indent after a line break (`cases/new_codestyle/annotation/AnnotationOnExpression.new.output`):

```kotlin
fun multipleAnnotations(): Int = @Suppress("A")
@Suppress("B")
compute()
```

* Destructuring declarations don't manage the closing bracket the same way as function calls, resulting in a weird formatting (`cases/new_codestyle/annotation/AnnotationOnExpression.new.output`):
```kotlin
val (
    @Anno("arg") first, @Anno("arg") second) = pair
```

Both these issues probably should be fixed in their corresponding visitors, and they do not directly depend on the assignment formatting. I'm mentioning it here just to (a) give context and (b) don't lose this.


core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangDeclarationFormatter.kt
```diff
@@ -159,7 +131,7 @@ internal class KotlinLangDeclarationFormatterImpl : DeclarationFormatterImpl() {
                     parameterList = component.parameterList,
                     typeConstraintList = null,
                     bodyExpression = component.bodyBlockExpression ?: component.bodyExpression,
-                    typeOrDelegationCall = component.returnTypeReference,
```
AbdullinAM, line 162: Replaced deprecated `returnTypeReference` with recommended `typeReference`, non-functional change

core/src/test/resources/cases/new_codestyle/annotation/AnnotationsOnTypes.new.output
```diff
@@ -33,8 +32,8 @@ fun @receiver:NotNull("arg") String.receiverOverflows(
 
 fun asExpression(x: Any) = x as @Ann String
 
-fun isExpression(x: Any) =
-    x is @Ann("arg") Map<String, List<Int>>
+fun isExpression(x: Any) = x is
```
kunyavskiy: This one looks bad.

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangCallFormatter.kt
```diff
@@ -74,14 +69,6 @@ internal class KotlinLangCallFormatterImpl : CallFormatterImpl() {
           format(expression.selectorExpression)
         }
       }
-      expression.isChainedScopingFunction &&
```
kunyavskiy: Looks like this change ened into a wrong MR

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangCallFormatter.kt
```diff
@@ -140,4 +169,23 @@ internal class KotlinLangCallFormatterImpl : CallFormatterImpl() {
       }
     }
   }
+
+  context(_: FormatterStateHolder)
+  override fun formatLambdaArguments(
```
kunyavskiy: Lambda parameters?

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangDeclarationFormatter.kt
```diff
@@ -172,4 +144,33 @@ internal class KotlinLangDeclarationFormatterImpl : DeclarationFormatterImpl() {
...
+      destructuringDeclaration: KtDestructuringDeclaration,
+  ) {
+    builder.sync(destructuringDeclaration)
+    val valOrVarKeyword = destructuringDeclaration.valOrVarKeyword
+    if (valOrVarKeyword != null) {
+      builder.token(valOrVarKeyword.text)
+      builder.space()
+    }
+    val hasTrailingComma = destructuringDeclaration.trailingComma != null
+    val openingDelimiter = destructuringDeclaration.lPar?.text ?: "("
```
kunyavskiy: How this can be null on green code? It feels like if it's null we would insert additional "(". Isn't it strange?

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangExpressionFormatter.kt
```diff
@@ -1,49 +1,85 @@
...
+      format(expression.left)
+      builder.space()
+      formatAssignmentLikeExpression(expression.right!!, expression.operationReference.text)
+      return
+    }
+
+    val allExpressions = expression.fullChain
+    format(allExpressions.first().left)
+    for ((index, currentExpression) in allExpressions.withIndex()) {
+      formatBinaryOperationToken(currentExpression, index == 0)
```
kunyavskiy: Please add named argument for boolean here.

core/src/test/resources/cases/new_codestyle/annotation/AnnotationOnExpression.new.output
```diff
@@ -9,10 +9,17 @@ private fun Window.animationQueue(): AnimationQueue =
 
 fun callFits(): Int = @Suppress("A") compute()
 
-fun callOverflows(): Int =
+fun callOverflows(): Int = @Suppress("UNCHECKED_CAST")
+compute(a)
```
kunyavskiy, line 13: Shouldn't there be a indent?

core/src/test/resources/cases/new_codestyle/annotation/AnnotationsOnTypes.new.output
kunyavskiy, line 35: Do we have some option to say that line break after `is/as` is somewhat last-resort, and we should do it only if there is no other option, and breaking before previous statement is better if helps?

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangDeclarationFormatter.kt
AbdullinAM: It probably can't ever be null, but `lPar` returns a nullable type. I think for clarity we can just remove the elvis

core/src/test/resources/cases/new_codestyle/annotation/AnnotationOnExpression.new.output
AbdullinAM: See "Known issues" section of the PR description. TLDR: there should be, but its probably not part of this PR

core/src/test/resources/cases/new_codestyle/annotation/AnnotationsOnTypes.new.output
AbdullinAM: There is no built-in option to express that. We can try to come up with workarounds.

For now the idea was: if you don't like how RHS is formatted, add a like break manually

core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/kotlinlang/KotlinLangDeclarationFormatter.kt
```diff
@@ -172,4 +144,33 @@ internal class KotlinLangDeclarationFormatterImpl : DeclarationFormatterImpl() {
 
     builder.guessToken(";")
   }
+
+  context(_: FormatterStateHolder)
+  override fun formatDestructuringDeclaration(
+      destructuringDeclaration: KtDestructuringDeclaration,
+  ) {
+    builder.sync(destructuringDeclaration)
+    val valOrVarKeyword = destructuringDeclaration.valOrVarKeyword
```
AbdullinAM, line 153: TODO: apply fix from #695

PR #693, opened 2026-09-09
[build] Set up publishing

AbdullinAM:
* Configure a publishing plugin
  * Use kotlinx libraries approach for publishing
* Configure native image plugin to create ready-to-be-published artifacts
  * The plan is to upload them to GH release pages

I will add new release instructions later, after the new setup is tested

### DIFF
--- a/buildSrc/src/main/kotlin/org/jetbrains/ktfmt/NativeImagePlugin.kt
+++ b/buildSrc/src/main/kotlin/org/jetbrains/ktfmt/NativeImagePlugin.kt
@@ -25,6 +25,7 @@ import org.gradle.api.file.RegularFile
 import org.gradle.api.plugins.JavaApplication
 import org.gradle.api.plugins.JavaPluginExtension
 import org.gradle.api.provider.Provider
+import org.gradle.api.tasks.Copy
 import org.gradle.api.tasks.Exec
 import org.gradle.api.tasks.TaskProvider
 import org.gradle.api.tasks.bundling.AbstractArchiveTask
@@ -32,6 +33,7 @@ import org.gradle.api.tasks.bundling.Compression
 import org.gradle.api.tasks.bundling.Tar
 import org.gradle.api.tasks.bundling.Zip
 import org.gradle.api.tasks.compile.JavaCompile
+import org.gradle.crypto.checksum.Checksum
 import org.gradle.jvm.tasks.Jar
 import org.gradle.kotlin.dsl.configure
 import org.gradle.kotlin.dsl.getByType
@@ -72,10 +74,11 @@ class NativeImagePlugin : Plugin<Project> {
     project.plugins.apply("application")
     project.plugins.apply("org.graalvm.buildtools.native")
     project.plugins.apply("signing")
+    project.plugins.apply("org.gradle.crypto.checksum")
 
     project.extensions.configure<JavaApplication> { mainClass.set(ENTRYPOINT) }
 
-    project.plugins.withId("java") { project.configureNativeImage() }
+    project.configureNativeImage()
   }
 
   private fun Project.configureNativeImage() {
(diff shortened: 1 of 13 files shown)


buildSrc/src/main/kotlin/org/jetbrains/ktfmt/utils.kt
```diff
@@ -0,0 +1,39 @@
...
+  LINUX("linux"),
+}
+
+internal enum class Arch(val archName: String) {
+  AARCH64("aarch64"),
+  X64("x86_64"),
+}
+
+internal fun Project.configurationProperty(name: String): Provider<String> {
+  return provider { findProperty(name) as? String ?: System.getenv(name) }
```
kunyavskiy: What about
providers.gradleProperty(name).orElse(providers.environmentVariable(name))
?

buildSrc/src/main/kotlin/org/jetbrains/ktfmt/NativeImagePlugin.kt
```diff
@@ -226,56 +208,93 @@ class NativeImagePlugin : Plugin<Project> {
...
+    }
+  }
+
+  private fun AbstractArchiveTask.configureNativeImageArchive() {
+    val archiveName = project.nativeImageArchiveBaseName
+    val archiveExtension = project.nativeImageArchiveExtension
+
+    description = "Packs the native image distribution into the publishable release archive"
+    dependsOn(project.tasks.named("nativeCompile"))
+    from(project.nativeImageExecutable) {
```
kunyavskiy: It's generally more ideomatic to not track dependency explictly, but use inputs-outputs for that. Can't we use from(project.tasks.named("nativeCompile")) instead?

buildSrc/src/main/kotlin/org/jetbrains/ktfmt/NativeImagePlugin.kt
```diff
@@ -226,56 +208,93 @@ class NativeImagePlugin : Plugin<Project> {
...
+      into(archiveName)
+      filePermissions { unix("rwxr-xr-x") }
+    }
+    archiveFileName.set("$archiveName.$archiveExtension")
+    destinationDirectory.set(project.layout.buildDirectory.dir("archive"))
+    isPreserveFileTimestamps = false
+    isReproducibleFileOrder = true
+  }
+
+  private fun Project.linesFromFile(fileName: String): List<String> {
```
kunyavskiy: I'm really surprised it works correctly with configuration cache. But at least this files should be inputs of the task.

buildSrc/src/main/kotlin/org/jetbrains/ktfmt/NativeImagePlugin.kt
```diff
@@ -226,56 +208,93 @@ class NativeImagePlugin : Plugin<Project> {
...
+
+    val key = signingKey.orNull
+    val password = signingPassword.orNull
+    if (key.isNullOrBlank() || password.isNullOrBlank()) return
+
+    val signing = extensions.getByType<SigningExtension>()
+    signing.useInMemoryPgpKeys(signingKeyId.orNull, key, password)
+    val archiveSignatures = signing.sign(archive.get())
+    artifacts.configure {
+      dependsOn(archiveSignatures)
```
kunyavskiy: That probably shouldn't be requied, provider tracks task dependency

buildSrc/src/main/kotlin/org/jetbrains/ktfmt/utils.kt
AbdullinAM: Based on the experimental evaluation: `providers.gradleProperty(name)` does not track properties injected by TC, `findProperty(name) as? String` --- does. 

We in theory can manually wire TC parameters via `-P...` and use `providers.gradleProperty(name)`. But for consistency with other "Kotlin Tools and Libraries" projects, I decided to use this setup

kunyavskiy (approved):

Issue #694, opened 2026-09-10
Annorations on destructuring declarations cause fail with FormattingException


AbdullinAM:
This (perfectly valid) code
```
fun foo() {
  @Suppress("unused")
  val (a, b) = 1 to 2

  @Suppress("unused")
  val [a, b] = 1 to 2
}
```

Causes following exception:
```
com.google.googlejavaformat.FormattingError: 2:3: error: expected token: '@'; generated val instead

	at com.google.googlejavaformat.OpsBuilder.token(OpsBuilder.java:342)
	at org.jetbrains.ktfmt.format.visitor.OpsUtilsKt.token(OpsUtils.kt:30)
	at org.jetbrains.ktfmt.format.visitor.OpsUtilsKt.token$default(OpsUtils.kt:29)
	at org.jetbrains.ktfmt.format.visitor.DeclarationFormatterImpl.formatDestructuringDeclaration(DeclarationFormatter.kt:387)
	at org.jetbrains.ktfmt.format.visitor.AbstractKotlinFormatter.visitDestructuringDeclaration(AbstractKotlinFormatter.kt:1021)
	at org.jetbrains.kotlin.psi.KtVisitorVoid.visitDestructuringDeclaration(KtVisitorVoid.java:530)
	at org.jetbrains.kotlin.psi.KtVisitorVoid.visitDestructuringDeclaration(KtVisitorVoid.java:10)
	at org.jetbrains.kotlin.psi.KtDestructuringDeclaration.accept(KtDestructuringDeclaration.java:42)
```

Root cause: annotations on destructuring declarations are parsed as modifier lists, and we do not handle them

PR #695, opened 2026-09-10
[format] Handle formatting of modifier lists on destructuring declara…

AbdullinAM:
…tions

#694 fixed


core/src/main/kotlin/org/jetbrains/ktfmt/format/visitor/DeclarationFormatter.kt
```diff
@@ -382,6 +382,13 @@ internal open class DeclarationFormatterImpl : DeclarationFormatter {
       destructuringDeclaration: KtDestructuringDeclaration,
   ) {
     builder.sync(destructuringDeclaration)
+    // The only modifiers that a desctructuring devlaration can have are modifiers.
```
kunyavskiy: Something is wrong with this statement. Not sure what you were trying to say.

PR #701, opened 2026-09-14
[kotlinlang] Preserve user-defined breaks in supertype lists

AbdullinAM:
The diff is currently broken because I haven't rebased on the latest `abdullin/initializers`. Will when the base branch will be finalised. For now we can review the formatting itself.

Open question here is what do we want to do with type constraints. Currently I haven't implemented anything additional, kept the default ktfmt format


core/src/test/resources/cases/new_codestyle/annotation/AnnotationOnExpression.new.output
```diff
@@ -55,6 +55,11 @@ val objectLiteral =
 
 class DelegationFits : I by @Anno impl
 
+class DelegationOverflows : InterfaceName by @Anno(
```
kunyavskiy, line 58: Why didn't we preserve line-break here?

